PR #293, opened 2013-11-27
Check SQL version using the product version rather than the year

subakva:
* Adds a new module for the version checking code
 * Changes the version check to look at the product rather than the year
 * Changes the adapter to use the new version checking method

This is more or less the same PR as https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/248

The caveats mentioned in the original PR still apply. Specifically, I still don't have a SQL Server instance to actually run the tests, and the meaning of database_year on the adapter is ambiguous.

Sorry about the lack of testing; I hope it still turns out to be useful.


metaskills:
We do not do any version checking in the v4.2.0 version

PR #302, opened 2014-02-11
Avoid retrieving all rows from information_schema tables

pisaruk:
We narrow the search space using a where clause in the query. It is
useful for databases where the number of tables are huge.


metaskills:
Could you open a new PR based off of the work done in master for v4.2. We now have a `Name` object that should really help our quoting and scheme needs. http://metaskills.net/2015/01/25/activerecord-sqlserver-v4.2.0---code-name-kantishna/

Issue #306, opened 2014-02-14
Incorrect SQL in Rails 4.0.1 when using Includes, Order and Limit


annaswims:
FinderTest#test_coerced_find_with_order_on_included_associations_with_construct_finder_sql_for_association_limiting_and_is_distinct does not pass

It generates different SQL in Rails 4.0.0 (good) and Rails 4.0.1-2 (failing).

``` ruby
Post.all.merge!(
      :includes => { :authors => :author_address }, 
      :order => 'author_addresses.id DESC ', 
      :limit => 2)
```

The SQL generated in Rails 4.0.0:

``` sql
EXEC sp_executesql N'SELECT TOP (2) [posts].id FROM [posts] LEFT OUTER JOIN [categorizations] ON [categorizations].[category_id] = [posts].[id] LEFT OUTER JOIN [authors] ON [authors].[id] = [categorizations].[author_id] LEFT OUTER JOIN [author_addresses] ON [author_addresses].[id] = [authors].[author_address_id] GROUP BY [posts].id ORDER BY MAX(author_addresses.id)'

EXEC sp_executesql N'SELECT [posts].[id] AS t0_r0, [posts].[author_id] AS t0_r1, [posts].[title] AS t0_r2, [posts].[body] AS t0_r3, [posts].[type] AS t0_r4, [posts].[comments_count] AS t0_r5, [posts].[taggings_count] AS t0_r6, [posts].[taggings_with_delete_all_count] AS t0_r7, [posts].[taggings_with_destroy_count] AS t0_r8, [posts].[tags_count] AS t0_r9, [posts].[tags_with_destroy_count] AS t0_r10, [posts].[tags_with_nullify_count] AS t0_r11, [authors].[id] AS t1_r0, [authors].[name] AS t1_r1, [authors].[author_address_id] AS t1_r2, [authors].[author_address_extra_id] AS t1_r3, [authors].[organization_id] AS t1_r4, [authors].[owned_essay_id] AS t1_r5, [author_addresses].[id] AS t2_r0 FROM [posts] LEFT OUTER JOIN [categorizations] ON [categorizations].[category_id] = [posts].[id] LEFT OUTER JOIN [authors] ON [authors].[id] = [categorizations].[author_id] LEFT OUTER JOIN [author_addresses] ON [author_addresses].[id] = [authors].[author_address_id] WHERE [posts].[id] IN (3, 2) ORDER BY author_addresses.id DESC'
```

[...]

PR #335, opened 2014-05-21
Fixing SQL generation for queries selecting COUNT(DISTINCT...) w/ ORDER_BY

nickveys:
Issue #306, and PR #329 broke compatibility with a couple of queries we were using.  Two regressions were introduced.

First, if the `COUNT(DISTINCT)` had an alias applied to it, the alias would be used in the `PARTITION BY` value expression, which was invalid SQL, [according to the Transact-SQL docs](http://msdn.microsoft.com/en-us/library/ms189461.aspx) it cannot use aliases.

Second, if the query (already having a COUNT(DISTINCT) and order clause) itself used a `select` call like this:

    Model.select('COUNT(DISTINCT)...')
         .select('something as x, something_else as y')
         ....
         .order(:something)

the outer select would lose some of the aliased columns due to the regex used when stripping out the aliases into `select_frags`.

Added a test to verify this, the issues are intertwined so it's just one test.

Thanks for the great gem!

### DIFF
--- a/lib/arel/visitors/sqlserver.rb
+++ b/lib/arel/visitors/sqlserver.rb
@@ -60,6 +60,9 @@ module Arel
         groups = core.groups
         orders = o.orders.uniq
 
+        # split out any projections that may have > 1 specified (comma-separated)
+        projections = projections.each_with_object(',').map(&:split).flatten
+
         select_frags = projections.map do |x|
           frag = projection_to_sql_remove_distinct(x, core, a)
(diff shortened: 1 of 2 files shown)


metaskills:
Closing this PR since it can not be applied to master now due to the 4.2 work. 

Perhaps you can help solve some of these remaining issues instead. https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/milestones/4-2-stable

PR #337, opened 2014-05-23
change sp_helptext to sp_columns for fixing "No view definition found" and "undefined method `match' for nil"

leckylao:
fixed "No view definition found" and NoMethodError - undefined method `match' for nil:NilClass: by changing sp_helptext to sp_columns


metaskills:
Thanks for the offer to help, but this will end up breaking some of our more detailed view support. The `sp_columns` does not return the view definition. We use `sp_helptext` as a secondary back when the views SQL is too long which is rare. Instead I have fixed the `views_real_column_name` method to be a bit more defensive when `sp_helptext` fails silently due to permission issues. Please find this in the upcoming v4.2.1 release.

PR #340, opened 2014-06-04
Invalid build of partition query throws SQL syntax error

kirrmann:
SQL syntax errors are thrown, because of invalid builded queries. Error comes from the use of the "AS" keyword in SELECT columns, what's not correct to put in the PARTITION statement.

There is also no need in specifying all the columns you want to get back from the PARTITION, because you want use them all anyway. That part was initially causing also some SQL errors.

I ran all tests and they were completed successfully.


metaskills:
Can you please try the latest v4.2.0.pre version of the adapter out. This is the only 4.x version that will be getting work done. http://metaskills.net/2015/01/25/activerecord-sqlserver-v4.2.0---code-name-kantishna/

PR #348, opened 2014-07-25
filter tables by current schema

l8nite:
We have a number of rails services talking to the same database and are using schema objects to logically partition the database. This is accomplished by setting the default schema for the independent database users that each service uses. Everything works great, except that when doing a db:schema:dump (for example, after a db:migrate) - the schema.rb file picks up all tables from all schemas.

This is a small change that filters the tables list by the session's current schema - there might be a better way to approach this - if you can think of one, let me know!

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -7,7 +7,7 @@ module ActiveRecord
         end
 
         def tables(table_type = 'BASE TABLE')
-          select_values "SELECT #{lowercase_schema_reflection_sql('TABLE_NAME')} FROM INFORMATION_SCHEMA.TABLES #{"WHERE TABLE_TYPE = '#{table_type}'" if table_type} ORDER BY TABLE_NAME", 'SCHEMA'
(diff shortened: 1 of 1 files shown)


metaskills:
I will look at getting this incorporated for the 4.2.0 release. Can you update this commit to work with our latest changes and tests?

metaskills:
Thanks for your contribution. While I think this a clean implementation it only solves your problem while opening up others. First, we have one test failing after this for a `table_exists?` check into another test schema. A small test, but it would have big consequences to those that want to rely on SQL Server's permission system handling which schemas a user is allowed to see. 

You see, things like `tables_exists?` are used lower down in the schema cache and column reflection. Unless we were to change all the top level schema statements to support an additional argument so we could pass in an explicit schema and bypass your `SCHEMA_NAME()` then the system breaks down.

metaskills:
FYI, I did try to think about this some more and we can not get it to work. The expected behavior of `tables` is to return all tables that user has permission to see. The `columns` method has a table_name argument and we are able to scope schema there using our new identifier object. This is where a fully qualified table name can shine for us. 

My suggestion is that you limit the permission server side as to shat schema/tables your users is allowed to see. Thus, you will get back a schema.rb as expected. Closing this ticket.

l8nite:
@metaskills thanks so much for taking the time to look at this. I had a feeling it was going to be hairier than proposed. I'll investigate figuring out how to restrict schema permissions in the manner you describe.

Issue #334, opened 2014-05-20
Database Tasks


agios:
I wanted to add support for database rake tasks, but I am not sure what should be done for the structure_dump and structure_load tasks. All the other tasks work fine already making it at least possible to run migrations etc, but for these I could not find a satisfactory cross-platform solution. 

For loading, previously sqlcmd was used, but I was thinking it should be possible to read in the file and the just send it to the driver - haven't tested it though, because it makes no sense without dumping.

Unfortunately, things are worse for dumping. Solutions so far had used smoscript, which requires windows to run properly, or a bunch of wine hacks. Another consideration was using SQL as described [here](http://stackoverflow.com/questions/21547/in-sql-server-how-do-i-generate-a-create-table-statement-for-a-given-table), but that's an even bigger hack and will probably be a pita to maintain.

Alternatively, those two specific tasks could raise an error indicating they are unsupported.

What do you think?

metaskills:
Closing, move talk PR.

PR #351, opened 2014-08-11
Add database tasks for Rails 4.2

agios:
These are db tasks, using FreeTDS for dump and load tasks, as discussed in #334

Other than dump, everything seems to work fine. Dump has some issues, due to FreeTDS defncopy bugs.

Firstly, defncopy had a few syntax errors when dumping, these are substituted in structure_dump, so at least the dump is valid and can be subsequently loaded.

However, defncopy also does not seem to dump foreign keys - this cannot be trivially fixed, without rewriting the schema dumping functionality ourselves. There is a warning stating this when you do a structure dump.

Also, if you want to try it and have views defined, it is possible that you will need to download FreeTDS and compile it with this patch https://gitorious.org/freetds/freetds/merge_requests/29

What do you think?

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -286,7 +286,11 @@ module ActiveRecord
         end
 
         def charset
-          select_value "SELECT SERVERPROPERTY('SqlCharSetName')"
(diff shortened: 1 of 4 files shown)


metaskills:
Whoa! Can you check to see if this works with the 4.2 work done in maser.

I will review this in a few days or so.

metaskills:
NOTE: FreeTDS `freetds-0.91.112` does have this patch.

metaskills:
Please read details in #380

PR #355, opened 2014-09-08
escape table name in change_column_null

krisl:
quote table name in change_column_null also


metaskills:
A better version that uses our newer name identifier object will be in the v4.2.1 release.

PR #361, opened 2014-11-14
Fix for offsets with Arel v5. Not dependent on deprecated Arel::Table#co...

chrisblatchley:
Arel::Table#columns is deprecated in v5. This is a fix to support #offset(). Also documented.

### DIFF
--- a/lib/arel/visitors/sqlserver.rb
+++ b/lib/arel/visitors/sqlserver.rb
@@ -425,12 +425,17 @@ module Arel
         end
       end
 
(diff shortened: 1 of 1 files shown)


metaskills:
Thanks! There has been major changes in the 4.2 adapter and this PR wil not work anymore. We likely want to focus on one 4.x version too, so please test that one out and re-open another if needed. More info. http://metaskills.net/2015/01/25/activerecord-sqlserver-v4.2.0---code-name-kantishna/

PR #365, opened 2014-11-25
Persistent connection errors during failover or with networking issues when executing stored procedures

rolftimmermans:
I am having issues with an application that makes extensive use of code such as:

    connection.execute_procedure(:fooBar, ...)
   
After a network or database node failure, errors are thrown in the application that subsequently do not go away. Also, the connection is reported as active after such an error has occurred:

    connection.active? #=> true

The errors often look like this:

    TinyTds::Error: DBPROCESS is dead or not enabled: EXEC fooBar 

When looking at the code, it seems to me that a call to `with_sqlserver_error_handling` is missing in this function:

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/604083bfd5fa8986e56ac0d54480a850d47b2cc0/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L84-L115

Replacing `@connection.execute` with `raw_connection_run` appears to fix the issue.


rolftimmermans:
Merge conflicts have been resolved.

test/cases/connection_test_sqlserver.rb
```diff
@@ -124,6 +124,31 @@ class ConnectionTestSQLServer < ActiveRecord::TestCase
       end
     end
 
+    should 'auto reconnect when setting is on for stored procedure' do
```
metaskills: Please change these to `it` statements. The mini-should compatibility layer is removed and these will fail in master soon.

test/cases/connection_test_sqlserver.rb
```diff
@@ -124,6 +124,31 @@ class ConnectionTestSQLServer < ActiveRecord::TestCase
...
+      end
+    end
+
+    should 'not auto reconnect on commit transaction' do
+      @connection.disconnect!
+      assert_raise(ActiveRecord::LostConnection) { @connection.commit_db_transaction }
+    end
+
+    should 'gracefully ignore lost connections on rollback transaction' do
+      @connection.disconnect!
```
metaskills: Please update these tests in general. They will not work in master now. For example, I have a custom `disconnect_raw_connection!` method since the adapter is now fully complain when calling `disconnect!` and hence clears out the raw connection ivar. Using `disconnect_raw_connection!` will address this.

metaskills:
I have made some comments that will need to be changed before this can be pulled. I wonder if you are trying to address a TinyTDS issue that has been resolved. In that projects master, I made this commit (https://github.com/rails-sqlserver/tiny_tds/commit/ac1bb7925e3240b42056b656cb70385759af5c78) and it will be out in the upcoming 0.7.0 release. Can you do a few things:

1) Try the latest TinyTDS master and confirm the issue still exists.
2) Update your tests to work with latest master and confirm they run.

```
$ bundle exec rake test TEST_FILES="test/cases/connection_test_sqlserver.rb"
```

That will run this file to QA.

metaskills:
There has been major changes in the 4.2 adapter and this PR wont work anymore. Please test the latest version out re-open a new issue if needed. http://metaskills.net/2015/01/25/activerecord-sqlserver-v4.2.0---code-name-kantishna/

PR #400, opened 2015-04-20
Views related fixes for structure dump and load

theangryangel:
This patch addresses 2 issues I have with structure dumping and loading:
- defncopy (at least the versions I have available) seems to truncate the definition of a view
- CREATE VIEW needs to be the first statement in a batch, this adds some very simple batch support by splitting the sql by GO and sending the commands separately

I'm not sure if these are the "correct" fixes, but it seems to be working for me at present. I'm happy to refactor as required.

As an aside I'm also having an issue with defncopy not pulling through auto-generating identities correctly. if you guys are happy with this pull request and don't tell me not to I'll be creating a new patch that completely removes defncopy from the equation to fix the identities problem.

### DIFF
--- a/lib/active_record/tasks/sqlserver_database_tasks.rb
+++ b/lib/active_record/tasks/sqlserver_database_tasks.rb
@@ -59,8 +59,6 @@ module ActiveRecord
         ]
         table_args = connection.tables.map { |t| Shellwords.escape(t) }
         command.concat(table_args)
-        view_args = connection.views.map { |v| Shellwords.escape(v) }
-        command.concat(view_args)
         raise 'Error dumping database' unless Kernel.system(command.join(' '))
         dump = File.read(filename)
         dump.gsub!(/^USE .*$\nGO\n/, '')                      # Strip db USE statements
@@ -69,10 +67,24 @@ module ActiveRecord
         dump.gsub!(/nvarchar\(-1\)/, 'nvarchar(max)')         # Fix nvarchar(-1) column defs
         dump.gsub!(/text\(\d+\)/, 'text')                     # Fix text(16) column defs
         File.open(filename, "w") { |file| file.puts dump }
+
+        # defncopy appears to truncate definition output in some circumstances
+        # Also create view needs to be the first operation in the batch.
+        File.open(filename, 'a') { |file|
+          connection.select_all("select definition, o.type from sys.objects as o join sys.sql_modules as m on m.object_id = o.object_id where o.type = 'V'").each do |row|
+            file.puts "\r\nGO\r\n#{row['definition']}"
+          end
+          file.puts "\r\nGO\r\n"
+        }
       end
 
       def structure_load(filename)
-        connection.execute File.read(filename)
+        structure = File.read(filename)
+        # Split by GO so that operations that must be in separate batches are in
+        # separate batches
+        structure.split(/^GO/).each { |s|
+          connection.execute s
+        }
       end
 
 
--- a/lib/active_record/tasks/sqlserver_database_tasks.rb
+++ b/lib/active_record/tasks/sqlserver_database_tasks.rb
@@ -71,8 +71,17 @@ module ActiveRecord
         # defncopy appears to truncate definition output in some circumstances
         # Also create view needs to be the first operation in the batch.
         File.open(filename, 'a') { |file|
-          connection.select_all("select definition, o.type from sys.objects as o join sys.sql_modules as m on m.object_id = o.object_id where o.type = 'V'").each do |row|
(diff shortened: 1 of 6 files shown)


theangryangel:
If the silence is simply because you guys are too busy, that's fine, and I apologise for bugging you. 

If it's more than you don't understand what or why I required this patch I can provide some additional information with practical examples. 

If you don't like the patch in it's current form, and I need alter how it works I can do so.

If it's a case of you don't want to accept it (with or without reason) I'd just rather know so I can plan for the future :)

metaskills:
@theangryangel Sorry for the silence, been super busy lately. Taking a look at this now.

lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -69,10 +67,24 @@ def structure_dump(filename)
         dump.gsub!(/nvarchar\(-1\)/, 'nvarchar(max)')         # Fix nvarchar(-1) column defs
         dump.gsub!(/text\(\d+\)/, 'text')                     # Fix text(16) column defs
         File.open(filename, "w") { |file| file.puts dump }
+
+        # defncopy appears to truncate definition output in some circumstances
+        # Also create view needs to be the first operation in the batch.
+        File.open(filename, 'a') { |file|
+          connection.select_all("select definition, o.type from sys.objects as o join sys.sql_modules as m on m.object_id = o.object_id where o.type = 'V'").each do |row|
+            file.puts "\r\nGO\r\n#{row['definition']}"
```
metaskills: Why use this SQL vs our existing `view_information` method? Does one work better than the other?

metaskills:
@theangryangel Made a few comments. This looks great and I can add it. Can you take a look at some of my comments first? Also, do you feel up to adding a test to the `test/cases/rake_test_sqlserver.rb` file? if not, I can help with that.

theangryangel:
No worries @metaskills - thanks for getting back to me, I didn't mean to come across as shitty - I'm finding a pattern of pull requests and patches getting ignored on various projects at the moment and I'd rather not spend what time I have on projects that won't accept them. I'll adjust my timescale expectations accordingly for this one :smile: 

The view_information I wasn't aware existed tbh. I did have a look but I must've missed it. I'll put aside some time over the next day or two and switch to that if it's just as good/better, and I'll add some tests at the same time as well :)

metaskills:
SWEET... can't wait... I very much appreciate the efforts too. A welcome change and I'll make sure we get this pulled in.

theangryangel:
I've updated the branch to use `view_definition` instead. I've added `routine_definition` and a few simple tests. I don't think I've missed anything that I should've added, but let me know if not and I'll ammend again.

On my test box all tests pass, with exception to a UUID one, but I've not touched that one, so I'm not sure if it was always broken.

metaskills:
Just writing to let you know I am still holding on this. I had a [recent trip to Microsoft](http://technology.customink.dev/blog/2015/10/19/microsofts-first-open-source-datacamp/) and I think we may have a cross platform native way to dump a DB that fixes all of this. If so, the adapter will be moving to it vs devncopy.

theangryangel:
The issue I have with defncopy is that it seems to generate the DDL itself (unless I'm being a complete idiot). This requires defncopy to be updated with each new column type (https://github.com/FreeTDS/freetds/blob/ea64a79533dad44c717d7b9764d6d97c8f925c58/src/apps/defncopy.c#L478). By utilizing SQL server to provide definitions of the objects itself, this mitigates the problem of keeping defncopy up to date.

That said, I believe this patch does need an additional update as I believe that I found large definitions were truncated with the original query I used. I'd need to double check though.

Just incase things have been forgotten over time, this patch also introduces alterations to the structure:load functionality, which splits the structure.sql by "GO" which allows for the file to have elements which isn't address by FreeTDS.

I'm happy to update as necessary, if it's worth my time.

metaskills:
I would hold off for a bit. Will repot back in a few days..

metaskills:
Re-opening. Not ignoring this, just have not got to it yet.

theangryangel:
Looking at the progress with #437 I assumed it was irrelevant, at least in-part, tbh :smile:

metaskills:
OK, after finally getting back into things, I would like to put the energy into #437. Apologies but please continue to use your branch of this work as needed. 

I will be doing some work in the 4-2-stable branch to get it running on SQL Server 2016 and running tests on TravisCI/CircleCI too. I believe finding a way to either allow development to happen on Windows/LINUX hosts which can run SQL Server and shell out to `sqlpackage` – or – provide instructions for POSIX systems to create some binstub that would do the same on a hosted VM.

PR #404, opened 2015-06-02
Lost engine context

pavels:
When you create new arel table and the mssql engine is secondary (main database is running on postgres for example), you loose context (queries than tries to run on primary connection and fails)

also on line 171 you have reference to @engine, which in Arel 6.0.0 (on my installation) is nil (and it also fails, but maybe i am missing something there)

### DIFF
--- a/lib/arel/visitors/sqlserver.rb
+++ b/lib/arel/visitors/sqlserver.rb
@@ -166,7 +166,7 @@ module Arel
         if Arel::Table === core.from
           core.from
         elsif Arel::Nodes::SqlLiteral === core.from
-          Arel::Table.new(core.from)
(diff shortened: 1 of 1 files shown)


metaskills:
Was just thinking of pulling this in, it does not break the tests, but I am not sure what it is doing. The latest Rails 4-x-stable compatible Arel version is 6.0.3. When I search the code, I can not see any `Transaction` constant. Where is that coming from? Was this a bug in 6.0.0 which is moot in 6.0.3?

metaskills:
Thanks... but closing this issue for now. If anyone has answers to my questions above, please ping me back and we can open this again.

PR #414, opened 2015-09-17
improve case comparision performace

faisalmansoor:
Case specific queries does not take in to account database column collation, therefore, the generated queries force an index scan. I noticed the problem while using uniqueness validator which specifically either does a case sensitive or case insensitive comparison (I didn't find a way to make it use default database casing). 

For a column with case insensitive collation, `SQL_Latin1_General_CP1_CI_AS`, uniqueness validator generates the following queries.

`validates :title, uniqueness: true` generates `SELECT  1 AS one FROM [articles] WHERE [articles].[title] = N''t1'' COLLATE Latin1_General_CS_AS_WS   ORDER BY [articles].[id] ASC OFFSET 0 ROWS FETCH NEXT 1 ROWS ONLY`

while `validates :title, uniqueness: {:case_sensitive => false}` generates `SELECT  1 AS one FROM [articles] WHERE LOWER([articles].[title]) = LOWER(N''t1'')  ORDER BY [articles].[id] ASC OFFSET 0 ROWS FETCH NEXT 1 ROWS ONLY` 

both causes index scan, for a large table, this cause a lot io.

MySQL adapter detects column collation and generates queries which does not cause table scan, I ported the same behavior to sqlserver-adapter in this PR. It will be great if you guys can provide some feedback, if this approach seems ok then I'll add some tests also.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -87,6 +87,22 @@ module ActiveRecord
           Arel::Nodes::Bin.new(node)
         end
 
+        def case_sensitive_comparison(table, attribute, column, value)
+          if column.case_sensitive?
+            table[attribute].eq(value)
+          else
+            super
+          end
+        end
+
+        def case_insensitive_comparison(table, attribute, column, value)
+          if column.case_sensitive?
+            super
+          else
+            table[attribute].eq(value)
+          end
+        end
+
         # === SQLServer Specific ======================================== #
 
         def execute_procedure(proc_name, *variables)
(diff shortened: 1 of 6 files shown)


lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -87,6 +87,22 @@ def case_sensitive_modifier(node, table_attribute)
...
+        end
+
+        def case_insensitive_comparison(table, attribute, column, value)
+          if column.case_sensitive?
+            super
+          else
+            table[attribute].eq(value)
+          end
+        end
+
```
faisalmansoor: These are copied from https://github.com/rails/rails/blob/602ec7a4a21846db74f0371eb27784f19f80e28c/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L768

metaskills:
Thanks so much for this work. I'll have master/v4.2.5 out in a few days or so.

PR #416, opened 2015-09-29
allow running sql batch with a given isolation level

faisalmansoor:
The following patch should allow running a batch of sql statements with a given isolation level.

### DIFF
--- a/lib/active_record/sqlserver_base.rb
+++ b/lib/active_record/sqlserver_base.rb
@@ -18,5 +18,24 @@ module ActiveRecord
       ConnectionAdapters::SQLServerAdapter.new(nil, logger, nil, config.merge(mode: mode))
     end
 
+    # Runs a batch of sql statements with an isolation_level
+    #
+    # For example:
+    #
+    #   ActiveRecord::Base.with_isolation_level(:read_uncommitted) do
+    #     Posts.where(id: 10).first
+    #   end
+    #
+    # This example will change the select query with :read_uncommitted isolation level
+    #
+    def self.with_isolation_level(isolation_level)
+      old_isolation_level_str = connection.user_options_isolation_level.to_s
+      isolation_level_str = connection.transaction_isolation_levels.fetch(isolation_level)
+      connection.execute "SET TRANSACTION ISOLATION LEVEL #{isolation_level_str}"
+      yield
+    ensure
+      connection.execute "SET TRANSACTION ISOLATION LEVEL #{old_isolation_level_str}"
+    end
+
   end
 end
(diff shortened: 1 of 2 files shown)


metaskills:
Thanks! But we support isolation levels using ActiveRecord's proper interface.

* Our `supports_transaction_isolation?` method. http://git.io/vcYFv
* Our database statements for isolation levels. http://git.io/vcYFa
* Our transaction classes for isolation support. http://git.io/vcYFH

Rails has had this since 4.0, and we finally removed our custom patches like this in favor of ActiveRecord's interface since the 4.2 adapter was published. Please read more here:

* Rails pull request for isolation levels. http://git.io/SEJDHQ
* Transaction isolation API. http://bit.ly/1sPwaM9

Hope that helps... gonna close this PR.

metaskills:
@faisalmansoor I was reading your test code for the PR too... if needed, could you give me more context on what you were trying to accomplish. I see your test code uses the core ActiveRecord `isolation` options for a transaction which would seem to indicate you knew about the links above. Thanks!

faisalmansoor:
Thanks Ken for the quick response.

I am trying to change isolation level within a transaction. 

E.g.
```
SET TRANSACTION ISOLATION LEVEL READ COMMITTED;
BEGIN TRANSACTION; -- Outer Transaction

SELECT dbo.CurrentIsolationLevel()

SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED; 
SELECT dbo.CurrentIsolationLevel();
SET TRANSACTION ISOLATION LEVEL READ COMMITTED;

SELECT dbo.CurrentIsolationLevel();
COMMIT;

--CREATE FUNCTION dbo.CurrentIsolationLevel()
--RETURNS VARCHAR(50)
--AS
--BEGIN
--    DECLARE @level VARCHAR(50);

--    SELECT @level =
--	   CASE transaction_isolation_level 
--		  WHEN 0 THEN 'Unspecified' 
--		  WHEN 1 THEN 'ReadUncommitted' 
--		  WHEN 2 THEN 'ReadCommitted' 
--		  WHEN 3 THEN 'Repeatable' 
--		  WHEN 4 THEN 'Serializable' 
--		  WHEN 5 THEN 'Snapshot' END
--	FROM sys.dm_exec_sessions 
--	where session_id = @@SPID;

--	RETURN @level;
--END
--GO;
```

I tried creating a nested transaction to change isolation level. But, active does not seems to support that.

```
  it 'allow executing sql statements with a given isolation_level in a nested transaction' do
    begin
      Ship.transaction(isolation: :read_committed) do
        Ship.transaction(isolation: :read_uncommitted) do
          connection.user_options_isolation_level.to_s.must_equal 'read uncommitted'
        end
      end
      connection.user_options_isolation_level.to_s.must_equal 'read committed'
    end
  end
```

this throw `ActiveRecord::TransactionIsolationError: cannot set isolation when joining a transaction
` exception. 

What do you think would be the best way to achieve this?

Thanks,
Faisal

metaskills:
Was going to merge this but it needs a rebase.

PR #419, opened 2015-10-14
Based on experience from #ossdatacamp - provides a more useful error …

jkingsbery:
…message when the username is anything other than user@short-name

I didn't understand that the username needed to be user@short-name, and the error messages I was getting when I did just `username` or `user@test.database.windows.net` weren't informative.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -277,6 +277,7 @@ module ActiveRecord
       end
 
       def dblib_connect(config)
+        raise "Username should be of the format user@short-name" unless config[:username] =~ /\S+@[A-z\-]+$/
         TinyTds::Client.new(
           dataserver: config[:dataserver],
           host: config[:host],


metaskills:
Interesting idea. Here are a few thoughts. 

* If we were going to raise an error, it should likely be `ArgumentError` or better yet, a subclass of that. Maybe `TinyTds::ArgumentError`. 
* More along that line, do you think this should be in the TinyTDS project? We already have things like this in there... see https://github.com/rails-sqlserver/tiny_tds/blob/master/lib/tiny_tds/client.rb#L63
* I wonder if this is exceptional? Maybe if we did move this in TinyTDS, we could just parse user name if opts :azure is true and do some lifting for the user?

jkingsbery:
Yeah, I actually like the approach of parsing the username out instead of raising an exception better.

I would leave it to your judgement whether this is better here or in TinyTds. Definitely makes sense to put validations in one place though, so since that's already in TinyTds, can put this there

metaskills:
I agree, if you are up for, please make a PR to TinyTDS where you have some type of private method like `azure_username(username)` that does all the work, perhaps using URI parse or something? Gonna close this one. Looking forward to it!

PR #426, opened 2015-11-23
Use fully qualified database name when retrieving column definitions …

jippeholwerda:
…to support linked servers.

We have to retrieve data from SQL Server 2008. Since that doesn't work with the latest activerecord-sqlserver-adapter and we do have the possibility of using a SQL Server 2012 instance, we've configured the 2012 database engine to be able to execute queries/commands against the 2008 instance. This can be achieved by adding the 2008 server as a ["linked server"](https://msdn.microsoft.com/en-us/library/ms188279(v=sql.110).aspx) in the 2012 instance.

However, querying a linked server requires using the fully qualified database name:
```
select * from [server].[database].[dbo].[table]
```

In the current master, using a linked server fails with the following error:
```TinyTds::Error: Invalid object name 'database.INFORMATION_SCHEMA.COLUMNS'```
when retrieving the table definition of a linked server.

This pull request should fix that by using the fully qualified database name.

An example of a model that uses a linked server is:
```
class LinkedServerTest < ActiveRecord::Base
  self.table_name = "[server].table.dbo.linked_server_test"
  
  # Setting the table alias used in Arel is necessary since otherwise the
  # fully qualified table name is used in the select_list of the select clause,
  # causing SQL Server to complain about the maximum number of prefixes.
  self.arel_table.table_alias = "linked_server_test"
end
```

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -221,7 +221,7 @@ module ActiveRecord
 
         def column_definitions(table_name)
           identifier  = SQLServer::Utils.extract_identifiers(table_name)
-          database    = "#{identifier.database_quoted}." if identifier.database_quoted
+          database    = identifier.fully_qualified_database_quoted
           view_exists = schema_cache.view_exists?(table_name)
           view_tblnm  = table_name_or_views_table_name(table_name) if view_exists
           sql = %{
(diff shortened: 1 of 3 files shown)


metaskills:
Brilliant side step solution for 2008 and Rails 4.2 and up! I just merged this in and published v4.2.6 of the adapter. Cheers!

metaskills:
@jippeholwerda Also... may I introduce you to my favorite minitest/rails gem :) https://github.com/metaskills/minitest-spec-rails

theangryangel:
Sorry for intruding, but I've got a question for @jippeholwerda - I had assumed that linked servers were passed SQL verbatim. Unfortunately I'm not in a position where I can test this until the end of the week and I'm dying to know the answer :sweat_smile: Are you saying that with a small 2012 (or newer instance) with a linked 2008 R2 server, I can issue FETCH and OFFSET queries (for example), and they're appropriately translated for 2008 R2?

metaskills:
That was my assumption too. I'd like to hear the answer @jippeholwerda

jippeholwerda:
@theangryangel That is correct. The 2012 engine correctly translates the FETCH and OFFSET queries so that 2008 understands.

theangryangel:
Shit. I wish I'd actually tried that rather than assumed it wouldn't. Would've saved me a bunch of trouble :laughing: Cheers for the update @jippeholwerda I appreciate it :heart:

jippeholwerda:
However, when testing I did run into an issue with INSERT statements. Apparently, specifying the Arel table alias adds it to the INSERT as well, resulting in a syntax error. Will have to dive further into that.

rept:
Great stuff.  And out-of-the box solution!

PR #427, opened 2015-11-24
Add linked server support to queries/commands

jippeholwerda:
Building on #426, this pull request fixes SELECT queries on linked servers. In addition, it is not necessary (and probably not desirable) to explicitly set the Arel table_alias for a model (causing other statements to break).

So, this should actually work:

```
class LinkedServerTest < ActiveRecord::Base
  self.table_name = "[server].table.dbo.linked_server_test"
end
```

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -216,7 +216,7 @@ module ActiveRecord
         end
 
         def sql_for_insert(sql, pk, id_value, sequence_name, binds)
-          sql = if pk && self.class.use_output_inserted
+          sql = if pk && self.class.use_output_inserted && !remote_server?
             quoted_pk = SQLServer::Utils.extract_identifiers(pk).quoted
             sql.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk}"
           else
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -220,7 +220,11 @@ module ActiveRecord
         end
 
         def column_definitions(table_name)
-          identifier  = SQLServer::Utils.extract_identifiers(table_name)
+          if remote_server?
+            identifier  = SQLServer::Utils.extract_identifiers("#{database_prefix}#{table_name}")
+          else
+            identifier  = SQLServer::Utils.extract_identifiers(table_name)
+          end
           database    = identifier.fully_qualified_database_quoted
           view_exists = schema_cache.view_exists?(table_name)
           view_tblnm  = table_name_or_views_table_name(table_name) if view_exists
--- a/lib/active_record/connection_adapters/sqlserver/utils.rb
+++ b/lib/active_record/connection_adapters/sqlserver/utils.rb
@@ -43,6 +43,10 @@ module ActiveRecord
             [server_quoted, database_quoted].compact.join(SEPARATOR)
           end
 
+          def fully_qualified?
+            parts.compact.size == 4
+          end
+
           def to_s
             quoted
           end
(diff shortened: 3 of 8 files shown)


jippeholwerda:
CI fails due to connection time-outs. Is there a way to rerun the CI tests?

jippeholwerda:
I've chosen a different way of making sure remote servers (linked servers) can be used in queries/commands. It is now a configuration option on the connection. This means it is no longer necessary to put the fully qualified server name in the table name.

Also, using this implementation it's not necessary anymore to handle all kinds of exceptional situations in ActiveRecord, related to the table name being fully qualified.

What do you think, @metaskills?

metaskills:
Looks interesting... what do you think about the failing tests?

jippeholwerda:
Yeah, not sure what happened. `rake test ONLY_SQLSERVER=1` didn't give any errors. I'll look into it.

jippeholwerda:
@metaskills Is there anything more I can do for this PR?

metaskills:
Nope, I'll get this merged in soon!

tommyalvarez:
@metaskills @jippeholwerda how do you go about setting up database.yml to connect to a LinkedServer database? I put the database statement as [LinkedServer].Database_Name but then i get:

TinyTds::Error: Could not locate entry in sysdatabases for database '[IG-DC02\SQLEXPRESS].Preactor_Igaltex'. No entry found with that name. Make sure that the name is entered correctly

metaskills:
There is no DB yaml setup. See #427 for details. I believe the feature is about using the schema prefix in table names via the AR model config.

tommyalvarez:
@metaskills I see, however if i put a fully qualified name “[linked_server].database_name.dbo.table_name” it raises an exception saying that the max level is 3. And if I try it like this: “[linked_server].database_name.[dbo.table_name]” says it’s invalid.

I spent all day trying to figure out how to connect a model to a table in a linked server.

jippeholwerda:
@tommyalvarez As I recall, in your database.yml you need to specify the `database_prefix` with something like `[linked.server].db.schema.`.

It might help to first see if you can actually execute a SQL query using a linked server in something like SQL Server management studio.

tommyalvarez:
@jippeholwerda after reviewing the commits in the pull request I managed to figure that out.
I did put in database.yml a database_prefix with the linked server database and schema.
However when performing something like RemoteModel.all the query produced, totally ignores the prefix.
I can confirm that by doing RemoteModel.connection.exec_query(“sql”) where sql includes the prefix works and performs the query

tommyalvarez:
@jippeholwerda okay progress so far:
After debugging this gem code i finally managed to set up the database_prefix in database.yml correctly. It was a key aspect that the prefix should end with ".". I now realized that in your comment it was `[linked_server].db.schema**.**`.  My bad, though very hard to guess, had to analyze the #parse_raw_name method in utils.rb file

So now at least, for a given model representing a table on a database of a linked server, column definition for the model is read correctly. However, i still can't query like `RemoteModel.all`. Doing so produces an sql like

`Select [table_name] from  [table_name]`

Of course the error returned is: `ActiveRecord::StatementInvalid Exception: TinyTds::Error: Invalid object name 'Table_Name'`. I'm sure i'm at the door of solving the issue. Do you by chance recall how to query remote tables in the linked server or how you need to configure the table_name in the model or any other thing in the database.yml? The one who is responsible for building the query is activerecord gem itself i think.

Some things i tried:
- In _database.yml_ if you set the attribute database to the name of the database in the linked_server it wont work as tiny_tds throws an error complaining that the database does not exist in sysdatabases, which is correct, if i query that system table, linked server databases won't show up.
- If you set to some database, for example _master_ in the server where the linked_server connection is established, it will not complain and everything works except of course querying the linked server database table through an active record model.
- If you change the table name to be fully qualified name, it complains that a table_name can only have 3 levels max, and fully qualified name is 4 (linked_server.database.schema.table).

@metaskills

metaskills:
I'll let @jippeholwerda reply. Linked servers are not my bag :) but open to helping get things in that make documentation better if needed.

PR #431, opened 2015-12-10
Use proper module for the sqlserver_connection method consistent with…

jippeholwerda:
… how other adapters work.

This also allows you to execute stored procedures from a class without having to inherit from `ActiveRecord::Base` (for instance, because there is no table to map to):

```
class NoTableModel
  extend ActiveRecord::ConnectionHandling
  include ActiveRecord::Core

  self.configurations = Rails.configuration.database_configuration
  self.establish_connection("#{Rails.env}_some_database".to_sym)

  def self.execute_procedure(*args)
    self.connection_pool.with_connection { |conn| conn.execute_procedure(*args) }
  end
end
```

### DIFF
--- a/lib/active_record/sqlserver_base.rb
+++ b/lib/active_record/sqlserver_base.rb
@@ -1,7 +1,6 @@
 module ActiveRecord
-  class Base
-
-    def self.sqlserver_connection(config) #:nodoc:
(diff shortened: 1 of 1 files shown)


metaskills:
Love it... this will be in the 4.2.7 soon.

Issue #447, opened 2016-02-15
Refine nvarchar quoting


RayBeaumont:
I know there has been a previous issue raised (#124) with regard to the nvarchar string prefix but I believe my problem is a bit different and I would like to raise an issue. 

**Background**
- The SQL Server database I am accessing is 2008 and cannot be updated
- All the varchar columns are `Latin1_General_CI_AS` and cannot be altered
- Since the database cannot be updated to 2012  I am having to use **activerecord-sqlserver-adapter v4.1**
- My database connection config contains `encoding: CP1252` which is a valid iconv encoding

**Problem**

The use of the `N` nvarchar prefix in the `where` clause causes a type mismatch when the query is executed and results in table scans which, on large queries, cripples the performance.  If I coerce my parameters to CP1252 and supply as an Array to the `where` method the `N` prefix is not applied :-). But it is still applied if supplied as a Hash :-(.

```
where("col = ?", 'value'.encode("CP1252"))
```

yields 

```
WHERE (col = ''value'')
```

which is cool. But 

```
where(col: 'value'.encode("CP1252"))
```

yields 

```
WHERE (col = N''value'')
```

which is not. 

This is because the current (v4.1) implementation of `Sqlserver::Quoting.quote` applies the prefix if a string column is given, regardless of the encoding of the value:

```
elsif value.is_utf8? || (column && column.type == :string)
  "#{quoted_string_prefix}'#{quote_string(value)}'"
```

Overriding `ActiveRecord::ConnectionAdapters::Sqlserver::Quoting.quoted_string_prefix` to remove the prefix entirely breaks the adapter.

**Solution**
One solution (below) is to amend the quote method to apply the prefix only if the value is utf8, whether the column is supplied or not. This seems like a rather drastic modification and whilst it seems to be working for me I'm concerned that it might be unwise?  I'd be interested in your thoughts and, if you think this a prudent solution, if it could be added to the 4.1 branch?.

```
elsif ((column && column.type == :string) && value.is_utf8?) ||
        (column.nil?  && value.is_utf8?)
  "#{quoted_string_prefix}'#{quote_string(value)}'"
```

metaskills:
Nice write up! I have to say, your solution looks clean. A few thoughts:

[...]

Issue #124, opened 2011-09-14
disabling N'...' quoting for strings


bpokosh:
I'm told that for our database using the following syntax is hurting performance:

SELECT FROM table WHERE field = N'...'

Because the field is a varchar (not unicode), being compared to a nvarchar value (unicode) is causing a performance hit.

Looking at the Quoting module it doesn't look like there is a way to override this behavior. Does it make sense to disable it when enable_default_unicode_types is false?

PR #461, opened 2016-04-07
Avoid the N' prefix on non-utf8 columns

marceloeloelo:
This is related to https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/447 and #124. 

As @RayBeaumont suggests:
> The use of the N nvarchar prefix in the where clause causes a type mismatch when the query is executed and results in table scans which, on large queries, cripples the performance.

My PR addresses removing the N' quoting when the column of the where clause is **not** a `NVARCHAR/NTEXT/NCHAR`.

We would need to keep an eye on the `quote` method once rails-5 goes out of beta since they are changing how column casting works there.

I'd like to add some tests but I've been unable to set up the environment for testing so far...

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/quoting.rb
+++ b/lib/active_record/connection_adapters/sqlserver/quoting.rb
@@ -47,21 +47,31 @@ module ActiveRecord
           end
         end
 
+        def quote(value, column = nil)
+          # records are quoted as their primary key
+          return value.quoted_id if value.respond_to?(:quoted_id)
+
+          if column
+            value = column.cast_type.type_cast_for_database(value)
+          end
+
+          _quote(value, column)
+        end
 
         private
 
-        def _quote(value)
+        def _quote(value, column = nil)
           case value
           when Type::Binary::Data
             "0x#{value.hex}"
           when String, ActiveSupport::Multibyte::Chars
-            if value.is_utf8?
-              "#{QUOTED_STRING_PREFIX}#{super}"
+            if (column && column.is_utf8?) || (column.nil?  && value.is_utf8?)
+              "#{QUOTED_STRING_PREFIX}#{super(value)}"
             else
-              super
+              super(value)
             end
           else
-            super
+            super(value)
           end
         end
 


marceloeloelo:
@metaskills this should be compatible with `rails 4.2.6`, check https://github.com/rails/rails/blob/v4.2.6/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L8

metaskills:
Thanks... but I was not in favor of adding `quote` back to 4-2-stable. I am almost certain that the condition described in 4-1-stable is better handled in our string type object. I'll have to look more on this later.

marceloeloelo:
The condition described in 4-1-stable also relies on the `column` data type which is not available in 4-2-stable within the Quoting module (that's the only reason I introduced the `quote` method).

I'm interested in seeing which other options we might have for this. The performance impact of adding the N' quote on `varchar` columns is really big.

marceloeloelo:
hey @metaskills, where you able to dig into this a little more? I'd be glad know you thoughts on how we could improve this. Thanks!

metaskills:
Not yet... started some TinyTDS testing for the 1.0rc1 of FreeTDS last night. But will get to here soon.

metaskills:
I made a few commits to master for the upcoming v4.2.11release that shows how to avoid N'' quoting for columns as needed. There has always been the #quoted_id feature of Rails which is a neat hack to use for this. This is the way I think people could handle char type columns in Rails 4.1.

```ruby
class CharValue
  def initialize(value)
    @value = value.to_s
  end
  def quoted
    "'#{@value}'"
  end
end
SSTestDatatypeMigration.where char_col: CharValue.new('Test')
SSTestDatatypeMigration.where varchar_col: CharValue.new('Test')
```

The new work in Rails 4.2 and upward (Rails 5) allows us to use the type objects. Tho somewhat janky, there is a pattern even used internal in Rails binary types to allow a special data object. I created a new char data type that you can pass to relations and get special quoting.

```ruby
data = ActiveRecord::Type::SQLServer::Char::Data
SSTestDatatypeMigration.where char_col: data.new('Test')
SSTestDatatypeMigration.where varchar_col: data.new('Test')
```

I know it is a bit manual, but I've been told that columns being used in the upper `#quote` method are indeed going away in Rails 5 and this is the way for these edge cases. Going to close this ticket, but still open to talking more if you want. Cheers.

metaskills:
Just did another small change. 

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/commit/a35835a11d88f4bef2a6e0bb2c7213d159b3a16b

Using AR's types in 4-2-stable upward allows us to return the Data wrapper which is then quoted properly. So now we can do this. 

```ruby
SSTestDatatypeMigration.where char_col: 'Test'
SSTestDatatypeMigration.where varchar_col: 'Test'
```

And get non-national quoting. Since there are some cases where ActiveRecord can not derive the column. I am keeping the `ActiveRecord::Type::SQLServer::Char::Data` type test/guard to always give people a forced choice when needed. Hope that helps.

marceloeloelo:
@metaskills that's a winner! Just tried it and it appears to be working good to me.

Thanks for taking the time of solving this. 

🏆

PR #471, opened 2016-05-13
Add to_s method to SQLServer::Type::Char::Data object

marceloeloelo:
@metaskills with the latest introduction of SQLServer::Type::Char::Data object one of the active record serializers that my app is using stopped working.

I chased the issue down to https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/type/string.rb#L35 where active record calls `value.to_s`.

Since `value` is of type `ActiveRecord::ConnectionAdapters::SQLServer::Type::Char::Data` it was not doing what is expected. 

Since the object doesn't have the `to_s` implementation it was returning a flat `"#<ActiveRecord::ConnectionAdapters::SQLServer::Type::Char::Data:0x000000044a0fe8>"` string, instead of returning the actual value from the database, breaking the serializer.

This PR adds just that, `to_s` implementation. Thoughts?

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/type/char.rb
+++ b/lib/active_record/connection_adapters/sqlserver/type/char.rb
@@ -24,6 +24,10 @@ module ActiveRecord
               "'#{@value}'"
             end
 
+            def to_s
+              @value
+            end
+
           end
 
         end


metaskills:
Good stuff! I added an `alias_method :to_str, :to_s` which is what ActiveRecord also does for the binary type's #to_s method. Will push this out in v4.2.13 soon.

marceloeloelo:
Thanks @metaskills! Do you have an idea on when v4.2.13 will be released? 😅

metaskills:
WHOOPS! Done now :)

marceloeloelo:
Thanks!

PR #478, opened 2016-06-01
Filter table constraints with matching table schema to column.

jonbell:
The following fix will correctly load column definitions across multiple schemas with similar table structures.  It is also significantly faster to load if using lots of schemas with similar structures/constraints.


metaskills:
Thanks, please try v4.2.16 with this change. I am also merging it into master and releaseing in a future v5.0.2 push.

Issue #487, opened 2016-06-21
changed_attributes returning inconsistent data


marceloeloelo:
Let's say I have an `Account` model which has some `varchar` columns. I'm seeing strange behavior when calling `changed_attributes`:

``` ruby
> Account
# => Account(ID: integer, Name: varchar, Address: string)

> account = Account.find(1)
# => #<Account ID: 1, Name: "ABCDE", Address: "123 Street">

> account.changed_attributes
# => {}

> account.save
# => true

> account.changed_attributes
# => { "Name" => "ABCDE" }
```

As you can see, there was no change to the `Name` attribute but it's still being included in the `changed_attributes` response _after_ the account object was saved.

@metaskills I can't explain this, do you have an idea on what might be going on? Have you experienced any similar issue with varchar?

Thanks

metaskills:
I have not, we certainly would not test the test if this were the case. Did you ever find out what the issue was? Going to close this but feedback still welcome.

PR #488, opened 2016-06-23
Type::Char::Data should be compared against it's string value

marceloeloelo:
Fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/487 

An optional different solution to the same problem could be:

``` ruby
module ActiveRecord
  module ConnectionAdapters
    module SQLServer
      module Type
        class String < ActiveRecord::Type::String

          def changed_in_place?(raw_old_value, new_value)
            if raw_old_value.is_a?(Char::Data) && new_value.is_a?(::String)
              raw_old_value.to_s != new_value
            else
              super
            end
          end
        end
      end
    end
  end
end
```

However, I think the PR's proposed code is more elegant and makes more sense. @metaskills thoughts?

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/type/char.rb
+++ b/lib/active_record/connection_adapters/sqlserver/type/char.rb
@@ -29,6 +29,10 @@ module ActiveRecord
             end
             alias_method :to_str, :to_s
 
+            def ==(other)
+              self.to_s == other.to_s
+            end
+            alias_method :eql?, :==
           end
 
         end


marceloeloelo:
@metaskills ping? 😺

metaskills:
Working on Rails v5 stuff for the next few weeks. Will round back here, I promise.

marceloeloelo:
Sounds good. Thanks for all your efforts!

ajs5710:
@marceloeloelo what issues are you trying to address?

marceloeloelo:
@ajs5710 the issue is described in #487

ajs5710:
@marceloeloelo ah I see now, thanks.  The reason I ask is because I have run into a similar problem and was trying to determine if your pull request fixes my issue.  

There's a table in our database that has a char() column for its id;  I added a uniqueness validator to the column and was met with an error about Type::Char::Data being compared against String.

The symptoms are a little different but I think that the root cause is the same.  I tried recreating your pull request in a config/initializer and it didn't seem to resolve the issue for me, though.  So, maybe not.  Thoughts?

marceloeloelo:
@ajs5710 that does sounds like the same issue. Not sure why your monkeypatch is not working though, sounds like it should be.

metaskills:
Thanks, please try v4.2.16 with this change.

Issue #402, opened 2015-05-13
TinyTds::Error: DBPROCESS is dead or not enabled


thomasfedb:
I'm receiving this error from a long-running worker process that periodically accesses a SQL Server database. Curious that some ActiveRecord machinery isn't attempting a reconnection, if that's the issue.

Full trace:
```
vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:308:in `execute': TinyTds::Error: DBPROCESS is dead or not enabled: EXEC sp_executesql N'SELECT [uvStudyLabTutors_CCGS].* FROM [uvStudyLabTutors_CCGS]' (ActiveRecord::StatementInvalid)
	from vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:308:in `raw_connection_run'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:299:in `_raw_select'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:295:in `block in raw_select'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/abstract_adapter.rb:466:in `block in log'
	from vendor/bundle/ruby/2.2.0/gems/activesupport-4.2.0/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/connection_adapters/abstract_adapter.rb:460:in `log'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:295:in `raw_select'
	from vendor/bundle/ruby/2.2.0/gems/activerecord-sqlserver-adapter-4.2.3/lib/active_record/connection_adapters/sqlserver/database_statements.rb:244:in `sp_executesql'
[...]

PR #495, opened 2016-07-15
Make connections reloadable

enriquez:
- For environments that set `config.caches_classes = false`, this will
  cause connections in the pool to reload between requests.
- Fixes an exception that is raised when a connection is idle for too
  long. See #402 

See relevant lines in ActiveRecord:
- https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railtie.rb#L154
- https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L447
- https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L415

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -160,6 +160,10 @@ module ActiveRecord
         do_execute 'IF @@TRANCOUNT > 0 ROLLBACK TRANSACTION'
       end
 
+      def requires_reloading?
+        true
+      end
+
       # === Abstract Adapter (Misc Support) =========================== #
 
       def tables_with_referential_integrity


metaskills:
Neat! So it seems only the SQLite3 adapter sets `requires_reloading?` to true. Why would this be good for us to always do this when other adapters like ours, for example Postgresql, do not?

enriquez:
I'm not sure about other adapters, but there seems to be a problem with SQL Server and idle connections (#402). This patch fixes it for me in the development environment by always getting a fresh connection.

metaskills:
> This patch fixes it for me in the development environment

I think that was my point too. 

* Is this right for production? Or is this setting not used in production env?
* Why does this patch work? Are we doing something wrong that this patch just sidesteps and the root fix is elsewhere?

Just trying to look to the other adapter to find where our fix may really be. A bit hesitant to patch this without a bit more understanding of a root cause. Especially since #402's proper fix was to wrap a long running job as gems like Sidekiq would do. Can you investigate more?

enriquez:
> Is this right for production? Or is this setting not used in production env?

This is only for environments that set `config.caches_classes = false`, so this shouldn't be used in production.

> Why does this patch work? Are we doing something wrong that this patch just sidesteps and the root fix is elsewhere?

You're right, this is probably sidestepping a root fix. Both this patch and the suggested fix for #402 will get a fresh connection instead of reusing an existing one from the connection pool. It's not ideal because of the overhead associated with getting a fresh connection every time. I think the long term fix would be to remove idle connections from the connection pool... [this is what they do in the .NET pooler](https://msdn.microsoft.com/en-us/library/8xx3tyca(v=vs.110).aspx):

> The connection pooler removes a connection from the pool after it has been idle for approximately 4-8 minutes...

ActiveRecord's connection pool doesn't have a way for us to do this, so we would probably have to fix it there. That seems like it would be a lot of work 😞

PR #502, opened 2016-08-11
Rails 5

metaskills:
This branch will be squashed and merged when complete. For now, we just got the adapter tests passing. I have yet to run the full ActiveRecord suite to see what else needs to be done but I suspect it will take a few more weeks till we are ready for a pre-release and testing.

screen shot 2016-08-10 at 8 02 29 pm


lib/active_record/connection_adapters/sqlserver/core_ext/explain.rb
```diff
@@ -4,11 +4,15 @@ module SQLServer
       module CoreExt
         module Explain
 
-          SQLSERVER_STATEMENT_PREFIX = 'EXEC sp_executesql '
-          SQLSERVER_PARAM_MATCHER = /@\d+ =/
+          SQLSERVER_STATEMENT_PREFIX = 'EXEC sp_executesql '.freeze
+          SQLSERVER_PARAM_MATCHER = /@\d+ = (.*)/.freeze
```
sgrif: Regex literals are always frozen in Ruby.

lib/active_record/connection_adapters/sqlserver/core_ext/explain.rb
```diff
@@ -4,11 +4,15 @@ module SQLServer
...
+          SQLSERVER_NATIONAL_STRING_MATCHER = /N'(.*)'/m.freeze
 
           def exec_explain(queries)
-            unprepared_queries = queries.map { |sql, bind| [unprepare_sqlserver_statement(sql), bind] }
+            unprepared_queries = queries.map do |sql_binds|
+              sql, binds = sql_binds
```
sgrif: You can inline this as `unprepared_queries = queries.map do |(sql, binds)|`

lib/active_record/connection_adapters/sqlserver/core_ext/explain.rb
metaskills: Thanks... keep it coming :)

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -19,11 +19,10 @@ def exec_query(sql, name = 'SQL', binds = [], sqlserver_options = {})
...
         end
 
-        def exec_insert(sql, name, binds, _pk = nil, _sequence_name = nil)
-          id_insert = binds_have_identity_column?(binds)
-          id_table  = table_name_from_binds(binds) if id_insert
-          if id_insert && id_table
-            with_identity_insert_enabled(id_table) { exec_query(sql, name, binds) }
+        def exec_insert(sql, name, binds, pk = nil, _sequence_name = nil)
+          id_insert_table_name = query_requires_identity_insert?(sql) if pk
+          if id_insert_table_name
```
sgrif: Do you think this would be more explicit as:

```ruby
if pk
  id_insert_table_name = query_rqeuires_identity_insert?(sql)
  ...
end
``

**Comment by sgrif** [comment]
#RandomUnsolicedCodeReview

File: lib/active_record/connection_adapters/sqlserver/database_statements.rb (hunk 3 above)
Comment (BrandonRoehl) [inline lib/active_record/connection_adapters/sqlserver/database_statements.rb]: If you still want to check that id_insert_table_name is not nil you could also do it like
```
if pk && (id_insert_table_name = query_requires_identity_insert?(sql))
  ...
end
```
This would check for pk and if it passes do the assignment and check. Parentheses may not be needed but clearly show the process.

File: lib/active_record/connection_adapters/sqlserver/schema_statements.rb [hunk 4]
```diff
@@ -407,8 +419,13 @@ def lowercase_schema_reflection_sql(node)
 
         # === SQLServer Specific (View Reflection) ====================== #
 
+        def view_exists?(table_name)
+          identifier = SQLServer::Utils.extract_identifiers(table_name)
+          views.include? identifier.object
+        end
```
BrandonRoehl: Does the identifier need to be stored in memory?
```views.include? SQLServer::Utils.extract_identifiers(table_name).object```
For a one liner that doesn't stamp out memory. Since it is only used once.

BrandonRoehl:
Yay merge!

metaskills:
## NOTICE

Please keep in mind this is only the core adapter tests passing. I'll be making individual PRs as work progresses for other things. There is still much work to do.

PR #504, opened 2016-08-20
[Rails5] Ensure date/time columns are not nil on create/update.

metaskills:
Creating or updating records resets the attributes by mapping the hash using ActiveRecord's `Attribute#forgetting_assignment` method. This method essentiall mimics the reload by setting the attributes to FromDatabase using the FromUser value "for the database".

https://github.com/rails/rails/commit/07723c23

This is a problem for us because the `value_for_database` for date time objects are strings that can not be re-parsed due to SQL Server's time formats. For example, a date of August 18th 2016 would be formatted as '08-19-2016' and fail deserialize:

``` ruby
ActiveModel::Type::Date.new.deserialize('08-19-2016') # => nil
```

Both ActiveModel's date and time types rescue `Date.parse` methods of `new_date` and `new_time` with nil. It was suggested that we use Data classes for our time types and we may explore that after this commit which essentially prepends a new `forgetting_assignment` method that checks the type. If it is a SQL Server date/time, it will just convert that value object to a FromDatabase using the existing value.

cc @sgrif @dpulliam

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/core_ext/attribute.rb
+++ b/lib/active_record/connection_adapters/sqlserver/core_ext/attribute.rb
@@ -1,0 +1,22 @@
+require 'active_record/attribute'
+
+module ActiveRecord
+  class Attribute
+
+    SQLSERVER_DATE_TIME_TYPES = [
+      Type::SQLServer::Date,
+      Type::SQLServer::DateTime,
+      Type::SQLServer::Time
+    ].freeze
+
+    prepend Module.new {
+      def forgetting_assignment
+        case type
+        when *SQLSERVER_DATE_TIME_TYPES then with_value_from_database(value)
+        else super
+        end
+      end
+    }
+
+  end
+end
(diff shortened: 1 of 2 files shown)


sgrif:
This patch would completely break dirty. You should just use a newtype wrapper in `serialize` and handle it in the adapter and `deserialize`

metaskills:
Documenting current pains of the alternate approach. No judgement, just sharing thoughts.

* Relying on Date/Time string parsing in Rails/Ruby to properly parse the time formats sent to the Database requires us to do more work in deserialization to make this happen. A good example is our DateTimeOffset type which can not rely on ActiveModel's `fallback_string_to_time`.
* It feels like we are comping what is sent to the Database in SQL to what we expect to get back from the database.
* In general we are loosing the benefits - or being forced to write code - where TinyTDS does this work for us. It feels like ActiveRecord has a lot of responsibility here that the C extensions should really be doing.

metaskills:
* `fast_string_to_time` does not support nanoseconds or zones.

Issue #526, opened 2016-11-30
Forward :contained login option to TinyTds


KDGundermann:
The new `:contained` login option introduced in TinyTds with 
[09dc4dc](https://github.com/rails-sqlserver/tiny_tds/commit/09dc4dcc8504f10f353827509ac0ae33f61abd7b)
is ignored by activerecord-sqlserver-adapter

metaskills:
Please see #527

PR #527, opened 2016-11-30
Fix for #526 : Forward config[:contained] to TinyTds

KDGundermann:
Targetting 4-2-stable branch

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -289,9 +289,10 @@ module ActiveRecord
           login_timeout: config_login_timeout(config),
           timeout: config_timeout(config),
           encoding:  config_encoding(config),
-          azure: config[:azure]
+          azure: config[:azure],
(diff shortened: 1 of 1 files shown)


metaskills:
I'll get to these soon. I've just rebooting the Rails 5 finishing work but want to get the new Docker/CircleCI stuff setup like in TinyTDS last week.

metaskills:
Thanks @KDGundermann we now have this work in the latest 4.2 and 5.0 release. I did not merge this PR because I did not think the [`:contained` option](https://github.com/rails-sqlserver/tiny_tds#tinytdsclient-usage) is not synonymous with `:azure`. 

When the [feature was first proposed](https://github.com/rails-sqlserver/tiny_tds/pull/292) it was specifically for contained DBs which could be hosted or azure cloud. To that end, we only pass that option down to the TinyTds client and do not issue the Azure specific `SET` commands if you only have `:contained` set. So the recommendation is that if you are using Azure, please continue to use that option.

I do wonder if we should set the `use_database` method to return if `:contained` is used?

Issue #535, opened 2016-12-29
activerecord-sqlserver-adapter-3.2.14 with rails 3.2.22.5, problem in user options


nielsjansendk:
Activerecord-sqlserver-adapter throws the following error at me:

/gems/activerecord-sqlserver-adapter-3.2.14/lib/active_record/connection_adapters/sqlserver/database_statements.rb:139:in `block in user_options': undefined method `gsub' for ["textsize", "2147483647"]:Array (NoMethodError)

I have tried this on ruby-2.1.10, ruby-2.2.6  and  ruby-2.3.3, same result. 

My database.yml file looks like this

>development:
> adapter: sqlserver
> mode: dblib
> dataserver: my_server
> database: my_database
> username: my_username
> password: xxxxxx
> timeout: 0

My Gemfile looks like this:

>source 'https://rubygems.org'
>gem 'rails', '3.2.22.5'
>
>gem 'tiny_tds', '~> 1.1'
>gem "activerecord-sqlserver-adapter", "~> 3.2.14"
>
>group :assets do
>  gem 'sass-rails',   '~> 3.2.3'
>  gem 'coffee-rails', '~> 3.2.1'
>
>  gem 'uglifier', '>= 1.0.3'
>end
>
>gem 'jquery-rails'

The application has nothing else in it, to isolate the problem. The error is thrown when I attempt to start a console. FreeTDS and Tinytds have been tested and work just fine.

metaskills:
So this line here: https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/3-2-stable/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L139

Does your users have permissions to run `dbcc useroptions`? If not, it is needed.

nielsjansendk:
Yes, I can execute "dbcc useroptions" from irb using tinytds, same connection properties.  The reuslts are:

>{"Set Option"=>"textsize", "Value"=>"64512"}
>{"Set Option"=>"language", "Value"=>"us_english"}
>{"Set Option"=>"dateformat", "Value"=>"mdy"}
>{"Set Option"=>"datefirst", "Value"=>"7"}
>{"Set Option"=>"lock_timeout", "Value"=>"-1"}
>{"Set Option"=>"isolation level", "Value"=>"read committed snapshot"}

metaskills:
[...]

PR #540, opened 2017-01-06
fix: enhance user_options' parsing (#535)

arbesulo:
In some cases the values are returned as an array, and not as a hash.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -135,9 +135,17 @@ module ActiveRecord
         
         def user_options
           return {} if sqlserver_azure?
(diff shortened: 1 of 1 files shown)


metaskills:
> I can help with the tests. I noticed this patch for user options first landed in 4-0-stable. I think we should patch it for 2-3-stable up to 3-2-stable.

Address #535 and I'll prioritize this after the Rails v5 work.

teaearlgraycold:
Applying this patch to master allows this to work for me in Rails 5, thanks

kride:
Same issue when upgrading from 4.2.3 to 5.0.1. This patch would fix it.

metaskills:
FYI, I'll have this work done this week in various branches.

metaskills:
Just pushed v3.2.16, v4.2.17, and v5.0.3 with this change.

PR #539, opened 2017-01-05
Rails5 supports datetime with precision

metaskills:
* Automatically use datetime2 when precision is used.
* Integration of datetime2 with `type_to_sql` so schema types are right.
* New subclasses for Table and TableDefinition with overhauled column methods.
* Fix our time fractional helpers to make datetime/datetime2 under full tests pass.
* New coerced tests.

PR #542, opened 2017-01-07
[Rails5] Pass more attribute tests for date/time/zones.

metaskills:
Another chapter in an epic series of getting our custom date time types passing all tests. I feel like we are in a constant battle that has gone downhill since we have to have specific quoting for each datetime type/precision that typically follows `%m-%d-%Y` format. This has been complicated by the fact that as part of forgetting assignment we could not rely on ActiveModel/Record to pare these strings for us which does not support a) our formats and b) time zones. I am willing to admit that all this work could somehow be easier but I can not see it. Other PRs include:

* https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/528
* https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/534
* https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/537
* https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/539

Now that we only have a few dozen tests failing. I might try this freedom patch again to see if things really break dirty support and/or helps reduce the complexity.

* https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/504

cc @sgrif

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/type/datetime.rb
+++ b/lib/active_record/connection_adapters/sqlserver/type/datetime.rb
@@ -11,8 +11,9 @@ module ActiveRecord
           end
 
           def serialize(value)
-            return super unless value.acts_like?(:time)
-            datetime = super.to_s(:_sqlserver_datetime).tap do |v|
+            value = super
+            return value unless value.acts_like?(:time)
+            datetime = value.to_s(:_sqlserver_datetime).tap do |v|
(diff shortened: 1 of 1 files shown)


metaskills:
OK, is what it is... #504 is no help in reducing complexity. I think it is all normal for us or my lack to find the simple path.

PR #560, opened 2017-02-07
Missing parentheses around native_text_database_type

BenKanouse:
These missing parentheses were causing the string setting and text
setting to behave differently. I think that we really want the setting
to update the text data type when it is set.

Here is a simplified example of what was happening:
```
2.1.3 :006 > boolean = false
 => false
2.1.3 :007 > "good" || (boolean ? 'bad' : 'should never happen')
 => "good"
2.1.3 :008 > "good" || boolean ? 'bad' : 'should never happen'
 => "bad"
```

Test screenshots:
Red
screen shot 2017-02-07 at 4 21 43 pm

Green
screen shot 2017-02-07 at 4 22 42 pm

### DIFF
--- a/test/cases/adapter_test_sqlserver.rb
+++ b/test/cases/adapter_test_sqlserver.rb
@@ -214,6 +214,42 @@ class AdapterTestSqlserver < ActiveRecord::TestCase
 
     end
 
+    context 'testing #native_string_database_type configuration' do
+
+      should 'use non-unicode types when set to false' do
+        with_native_string_database_type('varchar') do
+          assert_equal 'varchar', @connection.native_string_database_type
+          assert_equal 'nvarchar(max)', @connection.native_text_database_type
+        end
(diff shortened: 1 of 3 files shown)


test/cases/adapter_test_sqlserver.rb
```diff
@@ -214,6 +214,42 @@ class AdapterTestSqlserver < ActiveRecord::TestCase
...
+    context 'testing #native_string_database_type configuration' do
+
+      should 'use non-unicode types when set to false' do
+        with_native_string_database_type('varchar') do
+          assert_equal 'varchar', @connection.native_string_database_type
+          assert_equal 'nvarchar(max)', @connection.native_text_database_type
+        end
+      end
+
+      should 'use unicode types when set to true' do
```
BenKanouse: Copy-past error bitmoji

PR #566, opened 2017-02-23
Add TinyTDS as a runtime dependency.

metaskills:


metaskills:
See https://github.com/rails/rails/pull/28115 for origin.

PR #574, opened 2017-03-04
Perf w/inserts. Check binds & use schema cache.

metaskills:
Fixes #572. Thanks @noelr.


metaskills:
Tests are running much faster.

screen shot 2017-03-04 at 9 29 26 am

I dare say if we use Travis' cache for the OpenSSL/FreeTDS install then this will get much faster.

PR #577, opened 2017-03-14
Enable supports_json? for 2016 and up with a little bit of sugar.

### LINKED ISSUE #485: Ensure `supports_json?` Works
See this rails core PR for details. https://github.com/rails/rails/pull/21110

metaskills:
Adds a simple JSON type for creating an underlying `nvarchar(max)` column to add JSON too. Since SQL Server does not have a true JSON type. This means supporting JSON from a Rails perspective is a simple as adding the proper SQL Server type to your model's column which has some form of underlying string type.

```ruby
create_table :users do |t|
  t.string :name, :email
  t.json :data # Creates a nvarchar(max) column.
 end

class Users < ActiveRecord::Base
  attribute :data, ActiveRecord::Type::SQLServer::Json.new
end
```

This allows you to save and return JSON data and query it as needed.

```ruby
User.create! name: 'DJ Ruby Rhod', data: nil
User.create! name: 'Ken Collins', data: { 'admin' => true, 'foo' => 'bar' }

admin = User.where("JSON_VALUE(data, '$.admin') = CAST(1 AS BIT)").first
admin.data['foo'] # => "bar"
```

Fixes #485.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,24 @@ create_table :in_memory_table, id: false,
 end
 ```
 
+* Enable supports_json? Fixes #577.
+
+```ruby
+create_table :users do |t|
+  t.string :name, :email
+  t.json :data # Creates a nvarchar(max) column.
+ end
+
+class Users < ActiveRecord::Base
+  attribute :data, ActiveRecord::Type::SQLServer::Json.new
+end
+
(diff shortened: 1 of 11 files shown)


meet-bhagdev:
@metaskills this is really cool. Would something similar work in Tiny_TDS?

metaskills:
Oh yea... TinyTDS is just the means to execute SQL. In reality this work is more about documentation than doing anything.

Issue #593, opened 2017-05-23
OUTPUT - INTO seems not to be working


saeohioalpha:
I have an issue with output inserted on adapter v5.0.7 & Azure SQL (compat lvl 130)

The ActiveModel User.save method is triggering this. 

Not sure if this is my fault or if I am not putting a setting in that I need. 

This looks like it is trying to execute the following without the INTO section (from the log)
> EXEC sp_executesql N'INSERT INTO [users] ([Email], [PassHash], [First], [Last]) OUTPUT INSERTED.[id] VALUES (@0, @1, @2, @3)', N'@0 nvarchar(50), @1 nvarchar(50), @2 nvarchar(50), @3 nvarchar(50)', @0 = N'9ff019157811e18093fddb04daf3b69f@d.com', @1 = N'6935838143505C09CBB8779514089ED41233567F', @2 = N'sample', @3 = N'user'^[[0m  [["Email", nil], ["PassHash", nil], ["First", nil], ["Last", nil]]

I have tested this as working on the SQL server through ssms:
`DECLARE @inserted_id table (id int);
INSERT INTO [users] ([Email], [PassHash], [First], [Last]) 
	OUTPUT INSERTED.[id] AS INT INTO @inserted_id
	VALUES (
		N'9ff019157811e18093fddb04daf3b69f@d.com'
		,N'6935838143505C09CBB8779514089ED41233567F'
		,N'sample'
		,N'user'
	);
SELECT * FROM @Inserted_id;`

Also this works from the commandline using TSQL
`Setting example_db as default database in login packet
1> DECLARE @inserted_id table (id int);
2> INSERT INTO [users] ([Email], [PassHash], [First], [Last])
3>      OUTPUT INSERTED.[id] AS INT INTO @inserted_id
4>      VALUES (
5>              N'9ff019157811e18093fddb04daf3b69f@d.com'
6>              ,N'6935838143505C09CBB8779514089ED41233567F'
7>              ,N'sample'
8>              ,N'user'
9>      );
10> SELECT * FROM @Inserted_id;
11> GO
id
716162
(1 row affected)`

I also checked in the web console:
>  ActiveRecord::ConnectionAdapters::SQLServerAdapter.use_output_inserted
=> true

metaskills:
[...]

PR #595, opened 2017-05-26
Added new option "use_output_inserted_into" that defaults to false.

saeohioalpha:
Added new insert condition to use the OUTPUT INSERTED INTO TSQL statement format when the use_output_inserted_into option is set to true.
If the option is false, it uses the existing TSQL statement

This is due to SQL Server requiring the OUTPUT INSERTED INTO when a DB Table has a trigger. In that case a simple OUTPUT INSERTED will not work and will throw an error.

> The target table 'T' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.

#593

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -198,7 +198,12 @@ module ActiveRecord
           end
           sql = if pk && self.class.use_output_inserted && !database_prefix_remote_server?
                   quoted_pk = SQLServer::Utils.extract_identifiers(pk).quoted
-                  sql.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk}"
+                  if !self.class.use_output_inserted_into
+                    sql.dup.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk}"
+                  else
+                    tempsql = sql.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk} INTO @inserted_id"
+                    "DECLARE @inserted_id table (id int);#{tempsql}; SELECT * FROM @inserted_id;"
+                  end
                 else
                   "#{sql}; SELECT CAST(SCOPE_IDENTITY() AS bigint) AS Ident"
                 end
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -198,7 +198,7 @@ module ActiveRecord
           end
           sql = if pk && self.class.use_output_inserted && !database_prefix_remote_server?
                   quoted_pk = SQLServer::Utils.extract_identifiers(pk).quoted
-                  "DECLARE @inserted_identity table (#{quoted_pk} bigint);" << (sql.dup.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk} INTO @inserted_identity;") << ";SELECT CAST(#{quoted_pk} AS bigint) FROM @inserted_identity;"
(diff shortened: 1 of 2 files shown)


metaskills:
Thanks, I finally got the Rails v5.1 work done. In your core issue, you stated the following:

> On a table with a trigger, you have to do a OUTPUT INSERTED.id INTO not just OUTPUT INSERTED.id

Why not use `OUTPUT INSERTED.id INTO` all the time vs adding another config?

metaskills:
```
DECLARE @inserted_id table (id int)
```

This looks like it has a few issues. First, it would be the `quoted_pk`, second bigint is the default primary key for Rails now and we want to cast (like we do elsewhere) in this method to that type. We can't assume int for everything.

lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -44,6 +44,7 @@ class SQLServerAdapter < AbstractAdapter
 
       cattr_accessor :cs_equality_operator, instance_accessor: false
       cattr_accessor :use_output_inserted, instance_accessor: false
+      cattr_accessor :use_output_inserted_into, instance_accessor: false
```
metaskills: Would like to avoid using this and assume `use_output_inserted` leverages INTO as our default now

metaskills: Would like to avoid using this and assume use_output_inserted leverages INTO as our default now

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -198,7 +198,12 @@ def sql_for_insert(sql, pk, id_value, sequence_name, binds)
           end
           sql = if pk && self.class.use_output_inserted && !database_prefix_remote_server?
                   quoted_pk = SQLServer::Utils.extract_identifiers(pk).quoted
-                  sql.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk}"
+                  if !self.class.use_output_inserted_into
+                    sql.dup.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk}"
+                  else
+                    tempsql = sql.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk} INTO @inserted_id"
+                    "DECLARE @inserted_id table (id int);#{tempsql}; SELECT * FROM @inserted_id;"
```
metaskills: * Assume `use_output_inserted` uses `INTO`. No need for two formats.
* Change `(id int)` to something like `CAST(#{quoted_pk} AS bigint)`

saeohioalpha:
That could definitely be done. But since TSQL supported both I figured there may be use cases I am unaware of when someone would not want to use the updated language and I made it with that idea.

saeohioalpha:
I removed the extra option for :use_output_inserted_into
I have combined the conditionals to use the INTO format.
Changed the identity to use bigint as its datatype.

metaskills:
Thanks @saeohioalpha, are you actively working on this to get the tests passing? Right now it looks to be bombing out on the first fixture insert.

sardaukar:
Is there a workaround for this problem for the time being?

saeohioalpha:
Whoops, just saw this needed attention, I thought it had been merged in. Ill take a look.

saeohioalpha:
So im looking at the appveyor error and its crashing on the initial migration. 

`TinyTds::Error: Error converting data type nvarchar to bigint.: EXEC sp_executesql N'DECLARE @inserted_identity table ([key] bigint);INSERT INTO [ar_internal_metadata] ([key], [value], [created_at], [updated_at]) OUTPUT INSERTED.[key] INTO @inserted_identity VALUES (@0, @1, @2, @3);SELECT CAST([key] AS bigint) FROM @inserted_identity;', N'@0 nvarchar(4000), @1 nvarchar(4000), @2 datetime, @3 datetime', @0 = N'environment', @1 = N'development', @2 = '06-05-2017 20:55:46.333', @3 = '06-05-2017 20:55:46.333'`

It looks like this (ar_internal_metadata) is the table rails stores the migration information in. I have no idea why activerecord-sqlserver-adapter is trying to get the identity for a key->value table. Shouldn't it not be doing an identity insert on this table?

metaskills:
Closing this in favor of #625. A few things that drove that implementation. 

* The test here failed for any table that did not use a biginit PK.
* The overhead for the tmp table should be opt in.
* The tmp table output inserted into should be per table configurable.

metaskills:
Released v5.1.2 with this and another fix.

Issue #588, opened 2017-05-08
Rails 5: belongs_to_required_by_default


michaelrigart:
Since Rails 5.0, a new framework default has been introduced: ```belongs_to_required_by_default```

This option is being set in the file ```config/initializers/new_framework_defaults.rb``` . 
Since this is being configured through ```Rails.application.config```, it does not get picked up when using activerecord-sqlserver-adapter. 

This is due to the initialisation order of gems / application code. Details can be found at: https://github.com/rails/rails/issues/23589

Rails maintainer outline to use ```ActiveSupport.on_load(:active_record)``` when extending AR functionality and not use ```ActiveRecord::Base.send :include``` 

I have noticed that ```ActiveRecord::Base.send :include``` on a few places in the code, which produces the behaviour where belongs_to is no longer required, even though the option is set to true.

metaskills:
Sorry for the delay. Finally got finished with Rails v5.1 work. Great info and context links here too! I also liked this one form your Rails issue https://github.com/plataformatec/devise/commit/c2c74b0a39238e7d997486814a1c8f75fdaf276f

I'll have work in 5.1 and 5.0 soon to fix this. Do you think fixing the 3 freedom patches on ActiveRecord using `ActiveRecord::Base.send :include` would be the complete fixes? Or should we write [this file](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/master/lib/activerecord-sqlserver-adapter.rb) to something like this too?

```ruby
ActiveSupport.on_load(:active_record) do
  require 'active_record/connection_adapters/sqlserver_adapter'
end
```

michaelrigart:
My personal preference would be fixing the 3 freedom patches. It is more "explicit" .

While on the other hand, requiring the library entrypoint in a lazy-load hook would mean that if ever in the future, other patches would be included, these will automatically be loaded by the Rails lazy-load hooks.

metaskills:
OK... master is all green now with the 5.1 work done. If you want to work on a PR, that'll be great. Otherwise, I'll get to this next week.

PR #598, opened 2017-05-31
Bugfix: Use `ActiveSupport.on_load` to hook into ActiveRecord

michaelrigart:
See #588 
Use `ActiveSupport.on_load` to hook into ActiveRecord to respect the loading order. Otherwise configuration parameters might not be picked up in time.


metaskills:
Thanks so much, this is in master now under version 5.1.1. Will wait a few days or so to see if any more features come in before pushing to Rubygems. Thanks again!

PR #600, opened 2017-06-16
dateformat to ISO 8601

YvesR:
Change date format for SQL server and Azure that date format will
always be ISO 8601 format.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -143,14 +143,6 @@ module ActiveRecord
           end
         end
 
-        def user_options_dateformat
-          if sqlserver_azure?
-            select_value 'SELECT [dateformat] FROM [sys].[syslanguages] WHERE [langid] = @@LANGID', 'SCHEMA'
-          else
-            user_options['dateformat']
-          end
-        end
-
         def user_options_isolation_level
           if sqlserver_azure?
             sql = %(SELECT CASE [transaction_isolation_level]
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -418,14 +418,12 @@ module ActiveRecord
       def configure_application_name ; end
 
       def initialize_dateformatter
-        @database_dateformat = user_options_dateformat
-        a, b, c = @database_dateformat.each_char.to_a
-        [a, b, c].each { |f| f.upcase! if f == 'y' }
-        dateformat = "%#{a}-%#{b}-%#{c}"
+        # ISO 8601 format
+        dateformat = "%Y-%m-%d"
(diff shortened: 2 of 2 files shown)


lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -418,14 +418,12 @@ def configure_connection ; end
       def configure_application_name ; end
 
       def initialize_dateformatter
-        @database_dateformat = user_options_dateformat
-        a, b, c = @database_dateformat.each_char.to_a
-        [a, b, c].each { |f| f.upcase! if f == 'y' }
-        dateformat = "%#{a}-%#{b}-%#{c}"
+        # ISO 8601 format
+        dateformat = "%Y-%m-%d"
```
metaskills: Prefer descriptive variable names vs comments :) but we really don't need either.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -143,14 +143,6 @@ def user_options
...
         end
 
-        def user_options_dateformat
-          if sqlserver_azure?
-            select_value 'SELECT [dateformat] FROM [sys].[syslanguages] WHERE [langid] = @@LANGID', 'SCHEMA'
-          else
-            user_options['dateformat']
-          end
-        end
-
```
metaskills: Feels good to get rid of this!

YvesR:
Hello Ken,

1. about change request: I did remove the comment and renamed the variable from dateformat to dateformat_iso8601. How do I update this pull request now? As I can only create a new pull request as far as I can see.

2. I checked the details log. I did not took care of the tests. After changing the code I did some tests with rails applications. But going into tests now I see where they fails.

But please clarify: I need to change the tests as well to make them run successful?
I can change the values from us english format to iso 8601 in the tests as well, but I am not sure if this is right to just change the tests like this.
Also in ``column_test_sqlserver.rb' the test fail in line 285, I wounder about this, because we do not set language where I changed the code and string cast like this depends on language we do set before.

YvesR:
Hello @metaskills,

before I modify the tests, can you take a look on my last comment? It looks suspicious to me changing the tests to make them "just work".

metaskills:
Will do this weekend.

YvesR:
Saw that you just set a release 4 days ago. Since this is not included, did you have the time to take a look at this?

stefanahman:
@metaskills When is soon? 😆

wpolicarpo:
Hi @YvesR, I am jumping in some old issues/PR now...

Do you still need this merged? Can you give me some context and maybe rebase your branch and resolve the conflicts?

YvesR:
@wpolicarpo Hi. thanks for looking up. Well this fix is still useful and it make sense to do this. I was struggling with tests as we have to remove failed tests to make this work, but I wasn’t in the position to decide this.

aleksandrs-ledovskis:
@aidanharan This was closed on principle that such feature is not necessary or due to the staleness of PR's code?

PR #601, opened 2017-06-19
Added 2017 to supported list

catks:
I just added 2017 to the Supported versions constant


metaskills:
Thanks! Just pushed v3.2.18

PR #609, opened 2017-07-17
Fix mangled times after save when default_timezone is set to local

mdotterer:
Time attributes are being mangled in the Active Record after filter `changes_internally_applied` because the new attribute is being based of the serialized version of the time, which is a string. The new column value comes from parsing this string, but the old parsing method does not  account for the `default_timezone` setting causing the the time to be mis-interpreted.


metaskills:
Thanks! I really like your implementation of the `fast_string_to_time` method. The tests concern me and feel out of place. It strikes me as odd that somehow some of ActiveRecord's `BasicsTest` did not fail without this fix. Specifically the `test_preserving_time_objects_with_local_time_conversion_to_default_timezone_utc` test. Going to merge this in tho. Again, thanks!

PR #617, opened 2017-08-22
Refactor `SchemaDumper` to make it possible to adapter specific customization

kamipo:
Refer https://github.com/rails/rails/pull/30337

cc @metaskills


wpolicarpo:
These changes were applied to master in #653. Thanks @kamipo.

PR #624, opened 2017-09-12
Add case insensitive comparison for better performance with CI collations

attilahorvath:
Right now, the adapter will use the default `case_insensitive_comparison` from ActiveRecord which performs a conversion to lower case no matter what ([see here](https://github.com/rails/rails/blob/f10a6467c5252afa73cbdecb7400667fdda27ada/activerecord/lib/active_record/connection_adapters/abstract_adapter.rb#L428)). This gets called e.g. using the uniqueness validator with `case_sensitive: false`.

This conversion is unneeded and can be a performance issue if the collation in use is case insensitive anyway.

This PR is intended to avoid this when possible.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -84,7 +84,7 @@ module ActiveRecord
         end
 
         def can_perform_case_insensitive_comparison_for?(column)
-          column.type == :string
+          column.type == :string && (!column.collation || column.case_sensitive?)
         end
         private :can_perform_case_insensitive_comparison_for?
 


attilahorvath:
Actually, it's easier to just extend the `can_perform_case_insensitive_comparison_for` method with the check for CI collation. I've updated the code.

attilahorvath:
Those tests with the `No column name was specified for column 1 of 'subquery_for_count'` errors are failing for me even if I run them against the current master, so they are not caused by this change.

metaskills:
Correct, I fixed those errors in master. I'm looking over this now to see if I can write a test for this too.

metaskills:
One aspect of this is we really should likely put `can_perform_case_insensitive_comparison_for?` into the `case_sensitive_comparison` method like in rails-core so that when we super up on the else condition we can answer this question correctly.

metaskills:
Nope, this is perfect.

PR #625, opened 2017-10-01
Identity Inserts with Triggers

metaskills:
The adapter uses `OUTPUT INSERTED` so that we can select any data type key, for example UUID tables. However, this poses a problem with tables that use triggers. The solution requires that we use a more complex insert statement which uses a temporary table to select the inserted identity. To use this format you must declare your table exempt from the simple output inserted style with the table name into a concurrent hash. Optionally, you can set the data type of the table's primary key to return.

```ruby
adapter = ActiveRecord::ConnectionAdapters::SQLServerAdapter

adapter.exclude_output_inserted_table_names['my_table_name'] = true

adapter.exclude_output_inserted_table_names['my_uuid_table_name'] = 'uniqueidentifier'
```

Fixes #595 #593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Fixed
 
 * Patched `Relation#build_count_subquery`. Fixes #613.
+* Inserts to tables with triggers using default `OUTPUT INSERTED` style. Fixes #595.
 
 
 ## v5.1.2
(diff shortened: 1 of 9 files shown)


tommyalvarez:
@metaskills what if primary key is composite (multiple columns). Can you please provide an example of how the line ```adapter.exclude_output_inserted_table_names['my_uuid_table_name'] = 'uniqueidentifier'``` should be with say... a model that has as a primary key (using the composite primary key gem), ["columnA", "columnB", "columnc"] ? Is it supported?

PR #626, opened 2017-10-20
Adding 2017 to the supported versions.

avcastoldi:
Just adding `2017` to the list of supported versions.


metaskills:
Thanks! I've pushed 4.1.8 to rubygems.

PR #635, opened 2018-01-23
Use the query lock hint when joining tables

aidanharan:
When you perform a query using locking and a left-join, the lock is used in both the FROM and LEFT OUTER JOIN portions of the query:

**ActiveRecord:** `Person.lock(true).left_joins(:readers).load`
**SQL:** `SELECT [people].* FROM [people] WITH(UPDLOCK) LEFT OUTER JOIN [readers] WITH(UPDLOCK) ON [readers].[person_id] = [people].[id]`

However, when you use the same query with a inner-join, the lock is not used in the INNER JOIN portion of the query:

**ActiveRecord:** `Person.lock(true).joins(:readers).load`
**SQL:** `SELECT [people].* FROM [people] WITH(UPDLOCK) INNER JOIN [readers] ON [readers].[person_id] = [people].[id]`

This change will apply the lock hint to the inner-joins. 

I came across this issue as we are trying to use NOLOCK hints on queries containing joins. 

**ActiveRecord:** `Person.lock('WITH(NOLOCK)').joins(:readers).load`  
**Before change SQL:** `SELECT [people].* FROM [people] WITH(NOLOCK) INNER JOIN [readers] ON [readers].[person_id] = [people].[id]`
**After change SQL:** `SELECT [people].* FROM [people] WITH(NOLOCK) INNER JOIN [readers] WITH(NOLOCK) ON [readers].[person_id] = [people].[id]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
-## v5.1.5
+## v5.1.6
 
 #### Added
 
 * Use lock hint when joining table in query.
+
+
+## v5.1.5
 
 #### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v5.1.5
 
+#### Added
+
+* Use lock hint when joining table in query.
+
 #### Fixed
 
 * Memoize `@@version` queries. Fixes #632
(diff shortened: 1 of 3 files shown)


metaskills:
Love the work here, thanks!

metaskills:
Just pushed v5.1.6 to rubygems. Again, thanks!!!

PR #639, opened 2018-02-25
Primary key should be lowercase if schema forced to lowercase

one-more-alex:
I faced with errors while "rake db:schema:dump" on existing database.
This caused code do not match column object by name. Then null caused silent exceptions.
And schema had no dump for tables where primary keys where mixed case.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -93,7 +93,7 @@ module ActiveRecord
           identifier = database_prefix_identifier(table_name)
           database = identifier.fully_qualified_database_quoted
           sql = %{
-            SELECT KCU.COLUMN_NAME AS [name]
+            SELECT #{lowercase_schema_reflection_sql('KCU.COLUMN_NAME')} AS [name]
             FROM #{database}.INFORMATION_SCHEMA.KEY_COLUMN_USAGE AS KCU
             LEFT OUTER JOIN #{database}.INFORMATION_SCHEMA.TABLE_CONSTRAINTS AS TC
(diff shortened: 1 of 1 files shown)


taylorthurlow:
It would be fantastic to get this merged.

mbedarff:
I was faced with the same issue calling the id method on an active record object. Since this pull request isn't merged yet i wrote a small module that fixes this issue in the meantime:
```
module ActiveRecord
  module ConnectionAdapters
    module SQLServer
      module SchemaStatements
        module PrimaryKeysWithLowerCaseReflection
          def primary_keys(table_name)
            primaries = super
            lowercase_schema_reflection ? primaries.map(&:downcase) : primaries
          end
        end
      end
    end

    SQLServerAdapter.send(:prepend, SQLServer::SchemaStatements::PrimaryKeysWithLowerCaseReflection)
  end
end
```

This module is prepended automatically as a core extension that is applied during an intializer in my rails app.

engineersmnky:
If the intention of `lowercase_schema_reflection` is really just for convention to allow lower case method names and that is all. Then the simplest patch I can come up with is leave this option turned off and then add the following to `ApplicationRecord`: 

    def self.inherited(subclass)
      super
      subclass.class_eval do
        attribute_names.each do |attr|
          alias_attribute attr.downcase, attr
        end
      end
    end

This seems to be the least breaking change while still allowing all method names to be accessed accordingly. Additionally this would allow for conversion to snake case as well. 

     def self.inherited(subclass)
      super
      subclass.class_eval do
        attribute_names.each do |attr|
          alias_attribute attr.gsub(/(?<!^|[A-Z])[A-Z]/){"_#$&"}.downcase, attr
        end
      end
    end

Now if you have columns like "SomeTableNameId" or even "SomeTableNameID" it will be converted to: 

     some_table_name_id

Issue #640, opened 2018-03-02
Locking query issue, will you accept pr? 


ab320012:
Hi, 

I've been having an issue with production timeouts, i am solving it by monkey patching with (nolock) onto a call to information schema, wondering if you will accept a PR that modifies in the following way: 
```ruby
module ActiveRecord
  module ConnectionAdapters
    module SQLServer
      module SchemaStatements
        def view_information(table_name)
          @view_information ||= {}
          @view_information[table_name] ||= begin
            identifier = SQLServer::Utils.extract_identifiers(table_name)
            view_info = select_one "SELECT * FROM INFORMATION_SCHEMA.VIEWS with (nolock) WHERE TABLE_NAME = #{quote(identifier.object)}", 'SCHEMA'
            if view_info
              view_info = view_info.with_indifferent_access
              if view_info[:VIEW_DEFINITION].blank? || view_info[:VIEW_DEFINITION].length == 4000
                view_info[:VIEW_DEFINITION] = begin
                  select_values("EXEC sp_helptext #{identifier.object_quoted}", 'SCHEMA').join
                rescue
                  warn "No view definition found, possible permissions problem.\nPlease run GRANT VIEW DEFINITION TO your_user;"
                  nil
                end
              end
            end
            view_info
          end
        end
      end
    end
  end
end
```

pinak1180:
@ab320012  any point on why this is closed, I mean to ask is I tried to apply the patch and at the certain rate it used to reduce some timeout, though I know no lock is not as good solution, still want to know some details. thanks

ab320012:
@pinak1180  Hi Pinak, no reason for closing this really other than no response + no time way to delve into my side of the bug (which other scripts are running against this db). Basically "SELECT * FROM INFORMATION_SCHEMA.VIEWS WHERE TABLE_NAME = 'VIEW'" would time out extremely regularly in production and dev. After i added with (nolock) to the query this stopped, thought it may be useful to other people with this issue.

pinak1180:
@metaskills Can you please shed some light on this?

metaskills:
I re-opned this because using (nolock) here seems like a great idea. However, these slow DB calls should not happen in production because ActiveRecord caches them. In theory, they could happen when new workers are spawned under things like Passenger, but technically not because the columns should be cached prior to forking. Please read up on this thread a little for another project I host where a user experienced the same.

https://github.com/customink/secondbase/issues/55

[...]

PR #642, opened 2018-03-13
added with (nolock) hint to information_schema.views

ab320012:
#640


ab320012:
@metaskills do you need me to do anything else for this PR?

thomasfedb:
@metaskills this seems like a reasonable PR.

wpolicarpo:
I agree this would be good to have. Thanks @ab320012.

Issue #317, opened 2014-03-13
rake db:schema:dump only creates create_table in schema.rb


donhill:
In my database I login with a user that has no tables. The tables are in the dbo schema. When I run rake db:schema:dump the schema.rb gets create but no column definitions.

I tried putting ActiveRecord::Base.table_name_prefix = 'dbo.' in my Rakefile but I still get the same results. I hacked the file SchemaStatements#column_definitions by hardcoding

AND columns.TABLE_SCHEMA = 'dbo' ORDER BY columns.ordinal_position

this fixed the issue.

Is there something I am missing to configure the adapter to use schema dbo to use in the rake tasks?

metaskills:
The `dbo.` prefix is not needed since SQL Server's authentication and permission level should identify the current users as the owner and hence schema reflection should just work. That said, we have made some signification changes in the 4.2 version of the adapter and I would be interested in knowing if this is still an issue for. If so, please open up a new GH issue with details on how to replicate. Note, our test system creates a `rails` user which is the `dbo` and everything works great. PPS, our running unit tests doc mentions too that we assign `sysadmin` role for this user too. Maybe something to investigate.

mshappe:
This never appears to have been reopened, but am seeing the same thing -- I'm trying to pull the schema from an existing database. Some tables have complete information; many just have create_table and nothing more!

mshappe:
(Using v4.2.4 of the gem).

mshappe:
One issue of note for my situation -- the same database has two different schemata -- `dbo.` and `ns.`

mshappe:
The workaround above (hardwiring a namespace) basically works -- I can do two runs -- one for `dbo` and one for `ns`, and just move `schema.rb` aside so it doesn't get overwritten. This is, however, not ideal :-) As it happens, I'm mainly using this gem to work with legacy data and move it out into a different database, but if I were trying to actually adapt Rails to really use this legacy database, this behaviour would be a show-stopper.

mshappe:
Actually, it all works perfectly if I simply tell it to not constrain by TABLE_SCHEMA at all.

In column_definitions, near the end of the big query, simply remove

```
              AND columns.TABLE_SCHEMA = #{identifier.schema.blank? ? 'schema_name()' : '@1'}
```

entirely.

sjmadsen:
[...]

PR #655, opened 2018-08-15
Support for multiple schema types when dumping schema; potential fix for #317

alexggordon:
This is what I got. PR for specs and questions. 

This PR allows `rake db:schema:dump` to dump the database schema for all microsoft table schemas. 

It does this through exposing a method that can be monkey patched in an initializer:

```
module ActiveRecord
  module ConnectionAdapters
    class SQLServerAdapter < AbstractAdapter

      def configure_connection
        raw_connection_do "SET TEXTSIZE #{64.megabytes}"
      end

      def configure_application_name
        "myapp_#{$$}_#{Thread.current.object_id}".to(29)
      end

    end

    module SQLServer
      module SchemaStatements
        def server_schema
          ['dbo', 'schema2', 'schema3']
        end
      end
    end
  end

end
```

A couple of questions I have, because I don't know the internals of sql server super well. 

1. Should the `dbo` schema always be present in the `information_schema.tables` query? 
2. Is there a way to put the alternative schemas in the database.yml file that you know of?
3. When writing to the schema.rb should we include the `[schema_name].` in the `create_table` definition?
3b. How do we create the table in the same schema it was dumped in during `rake db:schema:load`?

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_dumper.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_dumper.rb
@@ -30,8 +30,21 @@ module ActiveRecord
         def default_primary_key?(column)
           super && column.is_primary? && column.is_identity?
         end
-
       end
+    end
+  end
+
+  class SchemaDumper
+    private
+
+    def remove_prefix_and_suffix(table)
+      table = remove_schema(table)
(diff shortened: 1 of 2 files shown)


wpolicarpo:
@alexggordon can you:

1) Explain better what you want to achieve this this and maybe give an example
2) Rebase your branch if you wan to go ahead this this?

bf4:
Closing as stale

PR #661, opened 2018-10-04
Coerce ActiveRecord test as table name wasn't being escaped

aidanharan:
The following test is currently failing in https://circleci.com/gh/rails-sqlserver/activerecord-sqlserver-adapter/209 because of a mistake in the Rails tests. I have coerced the test and fixed it in this pull request.

The issue was that the table name ("[authors]") was being used in a regular expression but the "[" and "]" characters were not being escaped to be used in a regular expression.

```
Failure:
ActiveRecord::RelationTest#test_relation_merging_with_merged_symbol_joins_is_aliased [/opt/circleci/.rvm/gems/ruby-2.2.10/bundler/gems/rails-76110f05fd93/activerecord/test/cases/relation_test.rb:235]:
Should be aliasing the child INNER JOINs in query

bin/rails test opt/circleci/.rvm/gems/ruby-2.2.10/bundler/gems/rails-76110f05fd93/activerecord/test/cases/relation_test.rb:227
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -859,6 +859,10 @@ class ViewWithPrimaryKeyTest < ActiveRecord::TestCase
     assert_equal 'id', model.primary_key
   end
 end
+
+
+
+
 class ViewWithoutPrimaryKeyTest < ActiveRecord::TestCase
   # We have a few view tables. use includes vs equality.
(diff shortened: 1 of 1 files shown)


aidanharan:
Just found that this is duplicate of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/660

Issue #40, opened 2010-06-21
Еncoding is not configurable


Excelan:
If one needs to connect a database with the encoding that deffers from UTF8 he have to past Iconv methods inside _raw_ metods 
(usr/lib/ruby/gems/1.8/gems/activerecord-sqlserver-adapter-2.3/lib/active-record/connection_adapters/sqlserver_adapter.rb)

PR #666, opened 2018-12-04
Fix build problems for testing.

rhuanbarreto:
Pointing the file download to http will fix the problems with Job #40.3 in travis ci. FTP download is unstable.


rhuanbarreto:
This is not related to the tests. But some steps cannot be reproduced without installing tiny_tds.

Issue #631, opened 2017-11-25
Rounding error when quoting times with usec > 999499 (off by ~1 sec)


buehmann:
We discovered this in v4.2.18.

When quoted, times with microseconds in the range 999500..999999 are rounded down to the next full second (effectively the microseconds are removed) instead of rounded up.

```
irb(main):154:0> ApplicationRecord.connection.class
=> ActiveRecord::ConnectionAdapters::SQLServerAdapter
irb(main):155:0> ApplicationRecord.connection.class.const_get('VERSION')
=> "4.2.18"
irb(main):156:0> ApplicationRecord.connection.quote('2017-11-25 10:05:38.999499'.to_time)
=> "'11-25-2017 09:05:38.999'"
irb(main):157:0> ApplicationRecord.connection.quote('2017-11-25 10:05:38.999500'.to_time)
=> "'11-25-2017 09:05:38'"
irb(main):158:0> ApplicationRecord.connection.quote('2017-11-25 10:05:38.999999'.to_time)
=> "'11-25-2017 09:05:38'"
```
I expect `'11-25-2017 09:05:39'` in the last two examples (or at least `'11-25-2017 09:05:38.999'`; that's what v5.0.7 seems to do).

PR #667, opened 2018-12-18
Fix Time value precision calculation

roooodcastro:
From [MS SQL docs](https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-2017), the datetime type has a millisecond precision of 1/300th of a second, or more specifically:

> Rounded to increments of .000, .003, or .007 seconds

The current implementation tries to round values passed as `datetime` to this standard, however the implementation is giving wrong values that do not end in either 0, 3 or 7. For example, if we pass a date ending in, say, `.8666667` seconds, it will incorrectly round to `.865` instead of `.867`, which would be the correct rounding for SQL Server, and also the correct rounding anywhere.

What I did was to eliminate a couple of inversions that were being done without any good reason (in fact they only made the result stray further from the right value), and increase the precision of the divider being applied (to make the overall precision 1/300th of a second).

Now milliseconds ending in `9` are being rounded to `10`, or more precisely, to `0`. This means that a date of `2017-11-25 10:05:38.999499` should be rounded to `2017-11-25 10:05:39.000`, but was instead being rounded to `2017-11-25 10:05:38.000`. There's actually an [open issue](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/631) about this. The second part of this PR addresses this. We now detect if there's a "carry second", and add it to the final date, after manually changing the microseconds to the rounded value.

Overall, because of the removed inversion logic, this conversion does not add any extra execution time, and this can demonstrated with a benchmark (which I will soon add here).

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/type/time_value_fractional.rb
+++ b/lib/active_record/connection_adapters/sqlserver/type/time_value_fractional.rb
@@ -10,12 +10,16 @@ module ActiveRecord
           def apply_seconds_precision(value)
             return value if !value.respond_to?(fractional_property) || value.send(fractional_property).zero?
             value.change fractional_property => seconds_precision(value)
+
+            millis = seconds_precision(value).to_i
+            value = value.change fractional_property => millis % fractional_operator
+            value + millis / fractional_operator
           end
 
           def seconds_precision(value)
             return 0 if fractional_scale == 0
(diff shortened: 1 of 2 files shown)


wpolicarpo:
Thanks for your contribution. This was probably introduced by 67317b3.

Issue #582, opened 2017-04-12
change_column_null uses cached definition of column type


aditya87:
When running migrations that change the column type and then add a null constraint the column type will be set back to the original type. This seems to happen because sqlserver/schema_statements.rb:239 references the schema cache and line 243 then sets that type on the column.

This can be reproduced with this example migration:
```
class SqlTest < ActiveRecord::Migration[5.0]
  class WithReference < ActiveRecord::Base;
  end

  def change
    create_table :foreigns do |t|
    end

    create_table :with_references do |t|
      t.string :foreign_id
    end

    change_column :with_references, :foreign_id, :integer

    # This resets the cache and makes this work. If you uncomment this the migration will pass.
    # WithReference.reset_column_information
    change_column_null :with_references, :foreign_id, false

    add_foreign_key :with_references, :foreigns
  end
end
```

An argument could be made that adding the null constraint and changing the column type should happen in one operation. However, this issue even seems to happen if the operations are spread across multiple migrations, since the cache doesn't get reset.

This same caching issue likely impacts other DDL operations as well.

FYI.: The ActiveRecord Postgres adapter doesn't seem to have this issue, since it clears out the cache before every DDL statement runs.

metaskills:
> The ActiveRecord Postgres adapter doesn't seem to have this issue, since it clears out the cache before every DDL statement runs.

I've noticed that too. I think they call the clear method explicitly right?

aditya87:
Yes they do.

ajmurmann:
https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb:389

They do this for every DDL statement. Not doing this might lead to different DB schemas in different environments. The cache doesn't seem to get reset even between separate migrations. So migrations will behave differently depending on which ones get executed in the same deploy, since not only separate statements will pollute each other via the cache, but even separate migrations.

metaskills:
Thanks y'all! I just finished the Rails v5.1 work in master and getting to a few issues. In that work change column now has `clear_cache!` at the end. https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/commit/e1e3bd02f59aea815675c927e8c76008b9e70f0a That one was needed to pass a new test in Rails core. 

Do you think I need to do more here and call `clear_cache!` to more methods like `change_column_null` too. Would be happy to add that or take a PR.

aditya87:
[...]

PR #681, opened 2019-02-22
change_column_null should not clear other column attributes

lairtonlelis:
When using `change_column_null` in the following migration, the column's limit, precision and scale were being cleared.

https://github.com/lairtonlelis/sqlserver_adapter_issue/blob/master/db/migrate/20190222183019_create_clients.rb

Seems to be the same issue as #582.


wpolicarpo:
Thanks for the PR!

PR #694, opened 2019-04-16
Change to name the columns when querying for table info

rdubya:
This is a change that is needed for the jdbc driver to work with this code so that the result has 2 columns in it. I have been working on the activerecord-jdbcmssql-adapter gem so that the jruby/jdbc specific differences can be maintained by that gem. I will cross link trackers as they are created. This will be needed in all the stable branches unless its already been fixed in future ones. I'm just starting with Rails 5.0 support so that's why I pointed the change at the 5-0-stable branch.


wpolicarpo:
Could you explain better what's the original problem here?

rdubya:
The first file's changes make it so that binds are not processed if prepared statements are off since they aren't used anyways and it breaks if you use a driver that can't handle them being passed in that case.

The second change gives names to the columns because otherwise both columns have the same name. With the way its currently being executed, that doesn't matter because it processes the columns by index and doesn't deal with the names at all, but it does matter if the driver needs the columns to be unique and its a little clearer on what is going on when looking at the query if we give the columns names.

kares:
any chance to re-consider merging these changes, are they blocked by something?

PR #703, opened 2019-07-05
Fix concatenate syntax

jensnockert:
This fixes concat to generate `a + b`, previously it generated the standard SQL `a || b` which SQL Server does not support.


wpolicarpo:
Thanks for this @jensnockert.

Issue #707, opened 2019-09-18
Undefined Method Error Caused By Active?


bendangelo:
Using version 5.1.6, activerecord 5.1.7

I tracked down the error to these lines. Starting from here:
```
def active?
        return false unless @connection
        raw_connection_do 'SELECT 1'
        true
      rescue *connection_errors
        false
      end
```
Raw connection is then called:
```
def raw_connection_do(sql)
          case @connection_options[:mode]
          when :dblib
            @connection.execute(sql).do
          end
        ensure
          @update_sql = false
        end
```

I believe the error is caused because `@connection.execute(sql)` can potentially return false. This causes the `undefined method do for false class error`. But I have not confirmed it.

Stacktrace:
```
NoMethodError: undefined method `do' for false:FalseClass
/usr/local/bundle/gems/activerecord-sqlserver-adapter-5.1.6/lib/active_record/connection_adapters/sqlserver/database_statements.rb:284:in `raw_connection_do'
/usr/local/bundle/gems/activerecord-sqlserver-adapter-5.1.6/lib/active_record/connection_adapters/sqlserver_adapter.rb:155:in `active?'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract_adapter.rb:444:in `verify!'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract/connection_pool.rb:810:in `block in checkout_and_verify'
/usr/local/bundle/gems/activesupport-5.1.7/lib/active_support/callbacks.rb:131:in `run_callbacks'
/usr/local/bundle/gems/activesupport-5.1.7/lib/active_support/callbacks.rb:827:in `_run_checkout_callbacks'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract/connection_pool.rb:809:in `checkout_and_verify'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract/connection_pool.rb:502:in `checkout'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract/connection_pool.rb:376:in `connection'
/usr/local/bundle/gems/activerecord-5.1.7/lib/active_record/connection_adapters/abstract/connection_pool.rb:933:in `retrieve_connection'
[...]

PR #708, opened 2019-09-27
Fixes active? method exception

bendangelo:
For context. https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/707


aharpervc:
If this is a problem with TinyTDS, then we should almost certainly be making this change in that repo rather than at this higher level

bendangelo:
Ok thank you. Yes this Pr has other side effects. Will close

PR #710, opened 2019-10-13
6 0 dev

serkaniyigun:

PR #714, opened 2019-11-15
rails 6.0 dependency

dglaser:
Included #710 changes from @serkaniyigun 
Since rails 6.0 has been released, I changed the gem depency from rails ~>6.0 beta3 to ~> 6.0

Tests are failing: 
`TinyTds::Error: Cannot insert explicit value for identity column in table [tasks] when IDENTITY_INSERT is set to OFF`

This could be as simple as setting IDENTITY_INSERT to ON/OFF. Doesn't help with activerecord tests that fail. 
`before do`
`    connection.execute("SET IDENTITY_INSERT on [table] ON;")`
`   ...`
`after do `
`  connection.execute("SET IDENTITY_INSERT [table] OFF;") `
`  .... `

### DIFF
--- a/activerecord-sqlserver-adapter.gemspec
+++ b/activerecord-sqlserver-adapter.gemspec
@@ -16,6 +16,6 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
-  spec.add_dependency 'activerecord', '~> 6.0.0.beta3'
+  spec.add_dependency 'activerecord', '~> 6.0.0'
   spec.add_dependency 'tiny_tds'
 end
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -223,7 +223,7 @@ module ActiveRecord
 
         protected
 
-        def sql_for_insert(sql, pk, sequence_name, binds)
+        def sql_for_insert(sql, pk, binds)
           if pk.nil?
             table_name = query_requires_identity_insert?(sql)
             pk = primary_key(table_name)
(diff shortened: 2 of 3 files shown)


ttilberg:
While digging backwards for the identity issue, I'm finding that one potentially related issue is that `ActiveRecord::ConnectionAdapters::SQLServerColumn` doesn't seem to correctly report whether it is an identity column. I could see this kind of problem causing all sorts of interesting issues when you run the test suite.

```
=> #<ActiveRecord::ConnectionAdapters::SQLServerColumn:0x000055ae6b885ee0
 @collation=nil,
 @comment=nil,
 @default=nil,
 @default_function=nil,
 @name="id",
 @null=false,
 @sql_type_metadata=
  #<ActiveRecord::ConnectionAdapters::SQLServer::SqlTypeMetadata:0x000055ae6b886070
   @limit=8,
   @precision=nil,
   @scale=nil,
   @sql_type="bigint(8)",
   @sqlserver_options=
    {:sqlserver_options=>
      {:ordinal_position=>1, :is_primary=>true, :is_identity=>true}},
   @type=:integer>,
 @sqlserver_options={}>
```

Notice specifically that `@sqlserver_options` is an empty hash, but `@sql_type_metadata`  knows it's an identity column deeper down. The check for identity_required? uses that `@sqlserver_options` ivar. 

Looking at the file lib/active_record/connection_adapters/sqlserver_column.rb -- the initializer is explicitly setting `@sqlserver_options` to an empty hash, and seemingly never doing anything with it, however these accessors are relying on the data it's supposed to entail.

We can see this was recently introduced at https://github.com/dglaser/activerecord-sqlserver-adapter/commit/b4f558e6927a05d46fe92e00e7ae2bd9d250f2b9 to handle kwargs, but was never assigned in the body.

ttilberg:
In the test run I'm working on, I'm seeing a ton of errors related to

```
NoMethodError: undefined method `split' for nil:NilClass
```

Upon digging into the stacktraces, this seems to be getting caused by something quirky in Minitest, not necessarily the AR adapter. I left some information here: https://github.com/seattlerb/minitest/issues/831

dglaser:
The PR you submitted is close to reverting the code I included from PR #710. I feel it would be cleaner to close this PR.

ttilberg:
Yes, I think it might be a better starting place if the rest works as expected.

utkarsh2102:
@ttilberg, you might be interested in this: https://github.com/seattlerb/minitest/issues/831#issuecomment-599946612 :)

PR #716, opened 2019-12-14
Translate the connection timed out error

midnight-wonderer:
Hi there,  
We have connection issues in our production environment.  
It is supposed to be an easy fix but we cannot safely rescue the exception because it is billed as an `ActiveRecord::StatementInvalid` exception.

I add my exception translation where it is supposed to be. It is not feasible to emulate the same issue in tests so I simply don't add any test.

Hopes you guys could release a quick fix version soon.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -154,7 +154,7 @@ module ActiveRecord
       def supports_lazy_transactions?
         true
       end
-      
+
       def supports_in_memory_oltp?
         @version_year >= 2014
       end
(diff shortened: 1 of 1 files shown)


midnight-wonderer:
BTW the full message is `TinyTds::Error: Adaptive Server connection timed out`

wpolicarpo:
@midnight-wonderer can you rebase your branch and resolve the conflicts?

lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -333,6 +333,8 @@ def translate_exception(e, message)
           NoDatabaseError.new(message)
         when /data would be truncated/
           ValueTooLong.new(message)
+        when /connection timed out/
```
wpolicarpo, line 336: Better match the full message `Adaptive Server connection timed out`. What do you think?

midnight-wonderer: The question is whether we use `StatementInvalid` or `StatementTimeout` when we encounter `connection timed out` String.

In my opinion, when `/time(?:d )?out/i` shows up it is very likely that `StatementTimeout` would be more accurate than `StatementInvalid`.

It depends on how we prioritize if we prioritize translation, loosely matching the String should be better. On the other hand, if we prioritize known cases and knowledge, matching the String exactly would be a better choice.

As a maintainer, which one you prioritize?

wpolicarpo: Ok, makes sense. Let's leave as it is.

PR #720, opened 2020-01-13
quoted_date doesn't work for Type::DateTime

ablignaut:
I've fixed the logic. I would happily alter the unit tests but I couldn't see any for this class?


wpolicarpo:
Can you provide a minimal test case or reproducible script showing us the error?

PR #743, opened 2020-03-19
Set default time precision when registering time type

aidanharan:
Should set the default precision of the time type when registering it rather than defaulting it within the `ActiveRecord::ConnectionAdapters::SQLServer::Type::Time` class.

This is a refactoring of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/737


aidanharan:
@wpolicarpo Could you have a look at this?

Issue #722, opened 2020-01-26
Tests from Rails fail with `TinyTds::Error: Ambiguous column name 'id'`


lcreid:
A number of the Rails tests fail with `TinyTds::Error: Ambiguous column name 'id'` (or something similar). It appears that the tests were added to Rails 5.2 by: https://github.com/rails/rails/pull/35360, and https://github.com/rails/rails/pull/35441 to test a fix to an issue in the Rails code (https://github.com/rails/rails/issues/35359).

I believe this issue is related to #713. 

I'm happy to take a stab at a PR. I will report back if I get blocked on the PR.

lcreid:
I've opened https://github.com/rails/rails/pull/38337, and submitted https://github.com/rails/rails/pull/38338 to Rails as a proposed fix for this issue.

lcreid:
The PR to Rails was merged to master and backported. I haven't had a chance to test the MS SQL Server adapter against Rails since the merge. If anyone gets a chance, please post here if it worked or if there were problems. 

The Rails PR doesn't fix *all* the test failures, but it fixes the ambiguous column name errors.

wpolicarpo:
Closing this as per #745. Thanks @lcreid and @aidanharan.

PR #745, opened 2020-03-24
Quoted table names containing square brackets need to be regex escaped

aidanharan:
Fixes the failing `activerecord/test/cases/relations_test.rb` tests. The tests were failing because table names in the SQL Server adapter use square brackets. The table names were then being used in a regular expression without first being regex escaped causing the regex not to work as expected. This PR escapes the table name in the regex. 

Table name not suitable for regex:
```
connection.quote_table_name(table.name) 
=> "[comments]"
```

Table name suitable for regex:
```
Regexp.escape(connection.quote_table_name(table.name))
=> "\\[comments\\]"
```


aidanharan:
Just found that @lcreid has already fixed this issue in Rails https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/722

In this PR I've taken the method from Rails master and patched the adapter so that the relations tests pass for Rails 5.2.4.x

lcreid:
Thanks, @aidanharan . I was going to ask that the Rails fix be back-ported, but this is much faster.

PR #755, opened 2020-04-02
Rails 6: Add/remove foreign keys when truncating during tests

aidanharan:
MSSQL does not allow tables to be truncated if they have foreign key constraints. This is the cause of the `ActiveRecord::StatementInvalid: TinyTds::Error: Cannot truncate table 'author_addresses' because it is being referenced by a FOREIGN KEY constraint.` errors in the tests.

For the tests that involve truncating just one table I have removed the foreign key, performed the test and then restored the foreign key at the end.

I have coerced the test that truncates all tables as I would need to remove all foreign keys in all tables and then later restore them and I don't know an easy way of doing that.

References:
- https://stackoverflow.com/questions/253849/cannot-truncate-table-because-it-is-being-referenced-by-a-foreign-key-constraint
- https://stackoverflow.com/questions/483193/how-can-i-list-all-foreign-keys-referencing-a-given-table-in-sql-server

**Before**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/669760945?utm_medium=notification&utm_source=github_status
```
6728 runs, 18647 assertions, 53 failures, 34 errors, 25 skips
```

**After**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/670157194?utm_medium=notification&utm_source=github_status
```
6725 runs, 18638 assertions, 53 failures, 29 errors, 25 skips
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -51,6 +51,56 @@ module ActiveRecord
   end
 end
 
+module ActiveRecord
+  class AdapterTestWithoutTransaction < ActiveRecord::TestCase
+    # SQL Server does not allow truncation of tables that are referenced by foreign key
+    # constraints. So manually remove/add foreign keys in test.
+    coerce_tests! :test_truncate_tables, :test_truncate_tables_with_query_cache
+    def test_truncate_tables_coerced
+      # Remove foreign key constraint to allow truncation.
+      @connection.remove_foreign_key :authors, :author_addresses
+
+      assert_operator Post.count, :>, 0
+      assert_operator Author.count, :>, 0
+      assert_operator AuthorAddress.count, :>, 0
+
+      @connection.truncate_tables("author_addresses", "authors", "posts")
+
+      assert_equal 0, Post.count
+      assert_equal 0, Author.count
+      assert_equal 0, AuthorAddress.count
+    ensure
+      reset_fixtures("posts", "authors", "author_addresses")
+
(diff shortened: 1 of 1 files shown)


test/cases/coerced_tests.rb
```diff
@@ -51,6 +51,56 @@ def test_value_limit_violations_are_translated_to_specific_exception_coerced
...
 end
 
+module ActiveRecord
+  class AdapterTestWithoutTransaction < ActiveRecord::TestCase
+    # SQL Server does not allow truncation of tables that are referenced by foreign key
+    # constraints. So manually remove/add foreign keys in test.
+    coerce_tests! :test_truncate_tables, :test_truncate_tables_with_query_cache
+    def test_truncate_tables_coerced
+      # Remove foreign key constraint to allow truncation.
+      @connection.remove_foreign_key :authors, :author_addresses
```
wpolicarpo, line 61: Can't you use connection#disable_referential_integrity instead?

aidanharan: No using `connection#disable_referential_integrity` does not work. The `disable_referential_integrity` method calls `ALTER TABLE [table] NOCHECK CONSTRAINT ALL` on the database but as you can see in the https://stackoverflow.com/a/3843826 and https://dba.stackexchange.com/a/190074 disabling constraints still does not allow tables to be truncated that have foreign key constraints.

PR #759, opened 2020-04-07
Rails 6: Store table names on columns to generate identities

aidanharan:
Rails no longer stores table names on columns since https://github.com/rails/rails/pull/35890 However, for the MSSQL adapter we need to continue to store the table name so that identity values can be calculated. See https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/8b48e5ff5bdb3e95d38437d0433735d0cf6f6b98/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L115

**Before**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/672064166
```
6728 runs, 18676 assertions, 53 failures, 16 errors, 25 skips
```

**After**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/672458067
```
6728 runs, 18680 assertions, 53 failures, 13 errors, 25 skips
```

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_column.rb
+++ b/lib/active_record/connection_adapters/sqlserver_column.rb
@@ -2,9 +2,9 @@ module ActiveRecord
   module ConnectionAdapters
     class SQLServerColumn < Column
 
-      def initialize(name, default, sql_type_metadata = nil, null = true, table_name = nil, default_function = nil, collation = nil, comment = nil, sqlserver_options = {})
-        @sqlserver_options = sqlserver_options || {}
-        super(name, default, sql_type_metadata, null, default_function, collation: collation, comment: comment)
+      def initialize(name, default, sql_type_metadata = nil, null = true, default_function = nil, collation: nil, comment: nil, **sqlserver_options)
+        @sqlserver_options = sqlserver_options
+        super
       end
 
       def is_identity?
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/sqlserver_column.rb
```diff
@@ -1,9 +1,12 @@
 module ActiveRecord
   module ConnectionAdapters
     class SQLServerColumn < Column
+      attr_reader :table_name
 
       def initialize(name, default, sql_type_metadata = nil, null = true, table_name = nil, default_function = nil, collation = nil, comment = nil, sqlserver_options = {})
         @sqlserver_options = sqlserver_options || {}
+        @table_name = table_name
```
wpolicarpo, line 8: We could get `table_name` from `@sqlserver_options` instead to avoid distancing even more from the original `ActiveRecord` constructor.

You would just need to just need to update `ActiveRecord::ConnectionAdapters::SQLServer::SchemaStatements#columns` to also slice `table_name` and create a  `table_name` method here delegating to `@sqlserver_options` as we do for `is_primary?` for example.

aidanharan: Done

PR #775, opened 2020-04-15
Rails 6: SQL Server does not support 'CREATE TABLE IF NOT EXISTS [table_name]'.

aidanharan:
The `MigrationTest#test_create_table_with_if_not_exists_true` test is failing because SQL Server does not support `CREATE TABLE IF NOT EXISTS [tablename]`. 

SQL generated:
```sql
CREATE TABLE IF NOT EXISTS [testings] ([id] bigint NOT NULL IDENTITY(1,1) PRIMARY KEY, [foo] nvarchar(4000))
```

Which causes the error:
```
ActiveRecord::StatementInvalid: TinyTds::Error: Incorrect syntax near the keyword 'IF'.
```

The main database adapters (Postgresql, MySQL and SQLite) all support the `CREATE TABLE IF NOT EXISTS` syntax.

There are workarounds to get the same functionality in SQL Server but the SQL is a lot different than what ActiveRecord generates. So in this PR we generate a `NotImplementedError` if you try to generate a `CREATE TABLE IF NOT EXISTS [tablename]` query.

References:
- https://lonewolfonline.net/sql-server-create-table-if-not-exists/

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_creation.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_creation.rb
@@ -10,8 +10,10 @@ module ActiveRecord
             table_name = quote_table_name(o.temporary ? "##{o.name}" : o.name)
             query = o.as.respond_to?(:to_sql) ? o.as.to_sql : o.as
             projections, source = query.match(%r{SELECT\s+(.*)?\s+FROM\s+(.*)?}).captures
-            select_into = "SELECT #{projections} INTO #{table_name} FROM #{source}"
+            "SELECT #{projections} INTO #{table_name} FROM #{source}"
           else
+            raise NotImplementedError.new("SQL Server does not support 'CREATE TABLE IF NOT EXISTS [table_name]'") if o.if_not_exists
+
             o.instance_variable_set :@as, nil
             super
           end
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/sqlserver/schema_creation.rb
```diff
@@ -10,8 +10,10 @@ def visit_TableDefinition(o)
             table_name = quote_table_name(o.temporary ? "##{o.name}" : o.name)
             query = o.as.respond_to?(:to_sql) ? o.as.to_sql : o.as
             projections, source = query.match(%r{SELECT\s+(.*)?\s+FROM\s+(.*)?}).captures
-            select_into = "SELECT #{projections} INTO #{table_name} FROM #{source}"
+            "SELECT #{projections} INTO #{table_name} FROM #{source}"
           else
+            raise NotImplementedError.new("SQL Server does not support 'CREATE TABLE IF NOT EXISTS [table_name]'") if o.if_not_exists
```
wpolicarpo, line 15: We should try to be compliant with ActiveRecord public API and implement the `if_exists` if possible. I know this is going to require a lot of changes but if it is possible we should try to keep the adapter compatible with all other databases.

aidanharan:
Closing in favour of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/779

PR #776, opened 2020-04-16
Rails 6: Fix tests that fail because of unique NULL indexes

aidanharan:
SQL Server differs from PostgreSQL/MySQL/SQLite in how it handles `NULL` values in unique indexes. 

For example, if you had a table with first and last names and created a unique index on both columns then the following 2 records would violate the unique index on SQL Server. 
 
| First Name  | Last Name |
| ------------- | ------------- |
| Bob  | NULL  |
| Bob  | NULL  |

However, in PostgreSQL/MySQL/SQLite unique indexes do not consider the NULL values equal and so the 2 records can be inserted successfully. 

This meant that some Rails tests were failing because the records required for the tests could not be inserted using SQL Server. The tests themselves were not testing unique indexes and assumed the records could be inserted.

In this PR the unique indexes are removed before the Rails tests are then called. The unique indexes are then restored afterwards. 

The code to coerce tests has been updated so that the coerced Rails tests can be called by prepending `original_` to their method name. This will help prevent duplicating the Rails test code in our test cases.

**Before**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/675315237
```
6731 runs, 18715 assertions, 30 failures, 10 errors, 25 skips
```

**After**
https://travis-ci.org/github/rails-sqlserver/activerecord-sqlserver-adapter/jobs/677268504
```
6731 runs, 18712 assertions, 30 failures, 5 errors, 25 skips
```

References:
- https://www.postgresql.org/docs/current/indexes-unique.html
- https://www.sqlitetutorial.net/sqlite-unique-constraint/
- https://dev.mysql.com/doc/refman/5.7/en/create-index.html#create-index-unique

### DIFF
--- a/test/support/coerceable_test_sqlserver.rb
+++ b/test/support/coerceable_test_sqlserver.rb
@@ -19,21 +19,28 @@ module ARTest
         end
 
         def coerce_all_tests!
-          once = false
           instance_methods(false).each do |method|
             next unless method.to_s =~ /\Atest/
             undef_method(method)
-            once = true
           end
           STDOUT.puts "🙉 🙈 🙊  Undefined all tests: #{self.name}"
         end
 
         private
 
-        def coerced_test_warning(method)
-          method = instance_methods(false).select { |m| m =~ method } if method.is_a?(Regexp)
+        def coerced_test_warning(test_to_coerce)
+          if test_to_coerce.is_a?(Regexp)
+            method = instance_methods(false).select { |m| m =~ test_to_coerce }
+          else
+            method = test_to_coerce
+          end
+
           Array(method).each do |m|
-            result = undef_method(m) if m && method_defined?(m)
+            result = if m && method_defined?(m)
+                       alias_method("original_#{test_to_coerce.inspect.tr('/\:"', '')}", m)
+                       undef_method(m)
+                     end
+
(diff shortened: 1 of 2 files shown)


test/support/coerceable_test_sqlserver.rb
```diff
@@ -19,21 +19,28 @@ def coerce_tests!(*methods)
...
+            method = instance_methods(false).select { |m| m =~ test_to_coerce }
+          else
+            method = test_to_coerce
+          end
+
           Array(method).each do |m|
-            result = undef_method(m) if m && method_defined?(m)
+            if m && method_defined?(m)
+              alias_method("original_#{test_to_coerce.inspect.tr('/\:"', '')}", m)
+              result = undef_method(m)
```
wpolicarpo, line 41: I know ruby's lexical scope will define `result` for us but having it explicitly defined at the same level as it is used makes it easier to understand. Maybe something like this:

```ruby
result = if m && method_defined?(m)
           alias_method("original_#{test_to_coerce.inspect.tr('/\:"', '')}", m)
           undef_method(m)
         end
```

aidanharan: Done

PR #786, opened 2020-04-27
Fix `EagerLoadingTooManyIdsTest#test_preloading_too_many_ids`

kamipo:
Since https://github.com/rails/rails/pull/36074, `in_clause_length` will
no longer help the adapter to avoid complex query limitation, and will
be removed from Rails code base (https://github.com/rails/rails/pull/39057).


wpolicarpo:
Thanks :)

PR #788, opened 2020-04-28
Rails 6: Coerce test "an empty transaction does not raise if preventing writes"

YehudaGold:
Because tests automatically wrapped in transaction the transaction in the test become SavepointTransaction and SQL Server does not have `release_savepoint` query so the assert should be corrected to 1.


wpolicarpo:
Can you rebase your branch?

PR #793, opened 2020-04-28
Rails 6: Clean coerced tests

YehudaGold:
I run the tests suite without the `coerced tests.rb` file and found old coerced tests that been fix in active record.

I removed old tests and tried to update tests to match the original tests.

I added comments to tests.


aidanharan:
Looks good. Could you squash your commits?

PR #809, opened 2020-05-11
Rails 6: Coerce reaper test using fork

aidanharan:
After Rails 6.0.3 the `ActiveRecord::ConnectionAdapters::ReaperTest#test_connection_pool_starts_reaper_in_fork` test is wrapped in `if Process.respond_to?(:fork)` so no need to coerce the test after that.

https://github.com/rails/rails/pull/37789

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1483,3 +1483,15 @@ class ActiveRecordSchemaTest < ActiveRecord::TestCase
     original_test_has_primary_key
   end
 end
+
+
+
+
+module ActiveRecord
+  module ConnectionAdapters
+    class ReaperTest < ActiveRecord::TestCase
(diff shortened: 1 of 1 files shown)


test/cases/coerced_tests.rb
```diff
@@ -1483,3 +1483,15 @@ def test_has_primary_key_coerced
...
 end
+
+
+
+
+module ActiveRecord
+  module ConnectionAdapters
+    class ReaperTest < ActiveRecord::TestCase
+      # Coerce can be removed if Rails version > 6.0.3
+      coerce_tests! :test_connection_pool_starts_reaper_in_fork
```
wpolicarpo, line 1494: You could coerce only if `Process` does not respond to `:fork` as Rails did originally.

aidanharan: Done

PR #845, opened 2020-08-27
Lateral apply

### LINKED ISSUE #738: Lateral support using CROSS/OUTER APPLY?
I've been trying to implement support for Arel::Nodes::Lateral by doing a CROSS/OUTER APPLY

This is nowhere near ready to be merged upstream, (it now always does an outer apply, no tests). For my usecase it seems to work with no ill effects, but before I put in the effort to make a proper PR, I want to gauge interest, does this make sense, is it a good idea etc...?

The comment in the PR lacks quite a bit of context.

https://github.com/jensnockert/activerecord-sqlserver-adapter/pull/1

sdahlbac:
Implement lateral using CROSS/OUTER APPLY

I added tests for both variants, both testing the generated sql and producing the expected result when executing the generated queries.

The tests pass, and I've also tested my use case for using lateral and that also works as expected.

Fixes #738 

Note however that this is my first significant contribution, so if I missed something/done something in a non-optimal way then please do let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@
 - [#855](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/855) Add helpers to create/change/drop a schema.
 - [#857](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/857) Included WAITFOR as read query type.
 - [#865](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/865) Implemented optimizer hints.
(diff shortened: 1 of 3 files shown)


sdahlbac:
@wpolicarpo or anyone, any chance of getting around to review this soon?

wpolicarpo:
Can you rebase your branch and add a changelog entry for this change?

sdahlbac:
@wpolicarpo Rebased and added changelog entry, do not understand the unrelated? 3.0.1 failure

PR #851, opened 2021-01-17
Updated 'column_definitions_sql' to ensure that only primary key key constraints are queried for

### LINKED ISSUE #850: Schema dumping in the Rails 6.0.0 Release is Broken Given a Table With Multiple Unique Constraints
## Issue

The following commit changed how columns were fetched for a table: 92ff32a9c28a1b9d0643b9c2ee272263f7fdd604

The new functionality duplicates columns in my schema.rb file once per unique key constraint, resulting in my schema.rb file being unable to properly stand up a new database due to duplicate column definitions. 

When I manually run the old SQL from before the aforementioned commit against my development database, the tables that meet the following criteria only get one entry per column. After the patch, the new SQL returns N number of columns depending on the number of unique indexes on the table. 

## Expected behavior
Each column definition should only be output to schema.rb once

## Actual behavior
Column names are duplicated in schema.rb

## How to reproduce

1. Create a DB table with the following SQL
```SQL
CREATE TABLE dbo.examples (
	id int IDENTITY(1,1) NOT NULL,
	example int DEFAULT 0 NOT NULL,
	CONSTRAINT IX_EXAMPLE UNIQUE (example),
	CONSTRAINT PK_EXAMPLE PRIMARY KEY (id)
);
```
2. Run `rails db:schema:dump`
3. Your schema.rb file should now contain the following code:
```ruby
  create_table "examples", id: :integer, force: :cascade do |t|
    t.integer "example", default: 0, null: false
    t.integer "example", default: 0, null: false
    t.index ["example"], name: "IX_EXAMPLE", unique: true
  end
```

## Details

- **Rails version**: `6.0.3.4`
- **SQL Server adapter version**: `6.0.0`
- **TinyTDS version**: `2.1.3`
- **FreeTDS details**:

  ```
                            Version: freetds v1.1.24
        Sybase binary compatibility: no
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: no
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
  ```

BenMcH:
Fixes #850 

Upon investigation into this query, I discovered that the new query is looking at the table `sys.key_constraints` which includes both primary keys as well as unique constraints, which is not what we want. 

Originally, I tried adding this check to the `where` clause, but my schema dump was missing any table without a primary key. After adding this to the join, I was able to successfully dump my schema, build a test database, and run my specs successfully against the new test database. 

CC @wpolicarpo

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -472,6 +472,7 @@ module ActiveRecord
               AND c.default_object_id = d.object_id
             LEFT OUTER JOIN #{database}.sys.key_constraints k
               ON c.object_id = k.parent_object_id
+              AND k.type = 'PK'
             LEFT OUTER JOIN #{database}.sys.index_columns ic
               ON k.parent_object_id = ic.object_id
               AND k.unique_index_id = ic.index_id
--- a/test/cases/schema_dumper_test_sqlserver.rb
+++ b/test/cases/schema_dumper_test_sqlserver.rb
@@ -135,6 +135,12 @@ class SchemaDumperTestSQLServer < ActiveRecord::TestCase
     assert_line :name, type: "string", limit: nil, default: nil, collation: nil
   end
 
+  it "dumps field with unique key constraints only once" do
+    output = generate_schema_for_table "unique_key_dumped_table"
+
+    _(output.scan('t.integer "unique_field"').length).must_equal(1)
+  end
+
   private
 
   def generate_schema_for_table(*table_names)
--- a/test/schema/sqlserver_specific_schema.rb
+++ b/test/schema/sqlserver_specific_schema.rb
@@ -277,4 +277,14 @@ ActiveRecord::Schema.define do
       field_2 int NOT NULL PRIMARY KEY,
     )
   SCHEMATESTMULTIPLESCHEMA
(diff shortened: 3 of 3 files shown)


wpolicarpo:
Looks good at first glance, but I'd like to double check with a minimal reproducible script before merging.

Do you think we could write a test for this?

BenMcH:
I'll take a look at writing a test for this, yes!

BenMcH:
@wpolicarpo All pushed up :smile: 

I reverted my change while writing the test, so I am sure that the test will fail if there is another regression. All tests were passing locally

wpolicarpo:
Thanks for the contribution @BenMcH. I'll cut a new release soon.

BenMcH:
Thank you @wpolicarpo! We look forward to pulling in the update :smile:

BenMcH:
@wpolicarpo is this going to be released soon?

wpolicarpo:
Sorry for the delay here @BenMcH. I'll cut a new release today or tomorrow, I promise.

wpolicarpo:
Version 6.0.1 released with your fix @BenMcH.

BenMcH:
Thank you, @wpolicarpo!

PR #852, opened 2021-01-27
Updated the column name matchers to accept database and owner names

aidanharan:
Updated the column name matchers to accept database and owner names. If you are working with multiple databases on the same SQL Server DBMS then you could have fully expanded column names that look like `[TestDB].[dbo].[Planets].[name]`. Currently this column name would not be accepted as the `[TestDB].[dbo]` section of the column name is not handled by the current regex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Changed
 
-- ...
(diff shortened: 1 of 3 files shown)


wpolicarpo:
Can you rebase your branch and add a changelog entry for it?

PR #853, opened 2021-02-05
Rails 6.1 support

lk0001:
Adds Rails 6.1 support.
Changes:
- use `ActiveRecord::ConnectionAdapters::SchemaCreation` instead of `ActiveRecord::ConnectionAdapters::AbstractAdapter::SchemaCreation`
- accept 2 arguments in `SQLServerRealTransaction#initialize` instead of 3

I don't know if that's all, but basic testing didn't reveal other issues.

PR #854, opened 2021-02-22
Rails 6.1 support and bit type bug fix

Circuit8:
This expands on #853 by fixing the test suite.

We also discovered that if a column is a "bit" type, and it has a default value of `true`, we would get an error from active record, as it seems that in the current version it is expecting boolean columns to be cast to `"1"` for true values, rather than `true`. This PR fixes that aswell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 #### Fixed
 
 - [#872](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/872) Use native String#start_with
+- [#876](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/876) Use native String#end_with
+- [#873](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/873) Various fixes to get the tests running for Rails 6.1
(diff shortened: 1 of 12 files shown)


marciojg:
@wpolicarpo Can you help us doing a merge of this PR and create a new release?

wpolicarpo:
@Circuit8 can you please rebase your branch against `main`?

Circuit8:
@wpolicarpo sure thing, should be sorted now.

runephilosof-abtion:
@Circuit8 You did not rebase this branch onto main. You merged main into this branch.

From https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/861#discussion_r613280110
> Not sure if that was the best approach to solve that issue. The fact that you had to change a spec rings a bell that something might be wrong there.
>
> TinyTDS casts the bit data type to boolean, can't see why we should change it at the adapter level. That could lead to unexpected bugs/changes to already existing apps assuming boolean as true/false instead of 1/0.
>
> In any case #875 addressed the original issue already.

wpolicarpo:
The remaining changes in this PR were superseded by #875 so I'm closing this off.

PR #855, opened 2021-03-02
feat: add helpers to create/alter/drop a schema

brunohkbx:
## Motivation
The native PostgreSQLAdapter from rails implement the following functions:

[create_schema](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/SchemaStatements/create_schema)
[drop_schema](https://apidock.com/rails/v4.1.8/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter/SchemaStatements/drop_schema)

So this PR aims to provide a similar interface to work with MSSQL schemas. Besides these two functions, I also created another one:

`alter_schema`
It can be useful to transfer a table from a schema to another. Read [more](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-schema-transact-sql?view=sql-server-ver15)

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -281,6 +281,21 @@ module ActiveRecord
           SQLServer::SchemaDumper.create(self, options)
         end
 
+        def create_schema(schema_name, authorization = nil)
+          sql = "CREATE SCHEMA [#{schema_name}]"
+          sql += " AUTHORIZATION [#{authorization}]" if authorization
+
+          execute sql
+        end
+
+        def change_table_schema(schema_name, table_name)
+          execute "ALTER SCHEMA [#{schema_name}] TRANSFER [#{table_name}]"
+        end
+
+        def drop_schema(schema_name)
+          execute "DROP SCHEMA [#{schema_name}]"
+        end
+
         private
 
         def data_source_sql(name = nil, type: nil)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 #### Added
 
-- ...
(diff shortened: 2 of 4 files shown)


brunohkbx:
@wpolicarpo would you mind reviewing it? Thanks in advance

wpolicarpo:
Could you update the description and add some information about these methods? This isn't part of the abstract adapter in ActiveRecord, right? I had a quick look and looks like only postgresql adapter implements it, which makes sense as sqlite and mysql are not supported.

brunohkbx:
They are part of the `PostgreSQLAdapter`. I thought it could be useful to have them here as well.

Where should I add docs for these functions? Is the example that I added in the README good enough?

lib/active_record/connection_adapters/sqlserver/schema_statements.rb
```diff
@@ -281,6 +281,21 @@ def create_schema_dumper(options)
...
         end
 
+        def create_schema(schema_name, authorization = nil)
+          sql = "CREATE SCHEMA [#{schema_name}]"
+          sql += " AUTHORIZATION [#{authorization}]" if authorization
+
+          execute sql
+        end
+
+        def alter_schema(schema_name, table_name)
```
wpolicarpo, line 291: I'm not sure if `alter_schema` is the best name here.

How about `change_table_schema(table_name, schema_name)` like there's [`change_table_comment`](https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionAdapters/SchemaStatements/change_table_comment) for example?

wpolicarpo (commented):
Added a few comments. Can you also rebase your branch and add a changelog entry for your changes?

brunohkbx:
Done @wpolicarpo :+1:  Thank you!

PR #861, opened 2021-03-22
Fix Rails 6.1 database config

### LINKED ISSUE #841: Queries on empty array causes TinyTds to throw
## Issue
Given the following code `MyModel.where(id: [])`, TinyTds will raise the following error:

```
ActiveRecord::StatementInvalid (TinyTds::Error: The isnull function requires 2 argument(s).)
```

## Expected behavior
The adapter should return an empty collection

## Actual behavior
It throws

## How to reproduce

1. Query a table with an empty array

## Details

- **Rails version**: `6.0.3.2`
- **SQL Server adapter version**: `6.0.0`
- **TinyTDS version**: `2.1.3.pre`

joshleblanc:
This PR expands on #854.

In addition to #854, it fixes a problem with `rake db:create` failing to run, as well as fixes queries on arrays throwing.

As a happy accident, this also fixes #841

### DIFF
--- a/lib/active_record/tasks/sqlserver_database_tasks.rb
+++ b/lib/active_record/tasks/sqlserver_database_tasks.rb
@@ -13,13 +13,18 @@ module ActiveRecord
       delegate :connection, :establish_connection, :clear_active_connections!,
                to: ActiveRecord::Base
 
+      def self.using_database_configurations?
+        true
+      end
+
       def initialize(configuration)
         @configuration = configuration
+        @configuration_hash = @configuration.configuration_hash
       end
 
       def create(master_established = false)
         establish_master_connection unless master_established
-        connection.create_database configuration["database"], configuration.merge("collation" => default_collation)
+        connection.create_database configuration.database, configuration_hash.merge("collation" => default_collation)
         establish_connection configuration
       rescue ActiveRecord::StatementInvalid => e
         if /database .* already exists/i === e.message
@@ -31,7 +36,7 @@ module ActiveRecord
 
       def drop
         establish_master_connection
-        connection.drop_database configuration["database"]
+        connection.drop_database configuration.database
       end
 
       def charset
@@ -49,14 +54,14 @@ module ActiveRecord
       end
 
       def structure_dump(filename, extra_flags)
-        server_arg = "-S #{Shellwords.escape(configuration['host'])}"
-        server_arg += ":#{Shellwords.escape(configuration['port'])}" if configuration["port"]
+        server_arg = "-S #{Shellwords.escape(configuration_hash['host'])}"
+        server_arg += ":#{Shellwords.escape(configuration_hash['port'])}" if configuration_hash["port"]
         command = [
           "defncopy-ttds",
           server_arg,
-          "-D #{Shellwords.escape(configuration['database'])}",
-          "-U #{Shellwords.escape(configuration['username'])}",
-          "-P #{Shellwords.escape(configuration['password'])}",
+          "-D #{Shellwords.escape(configuration_hash['database'])}",
+          "-U #{Shellwords.escape(configuration_hash['username'])}",
+          "-P #{Shellwords.escape(configuration_hash['password'])}",
           "-o #{Shellwords.escape(filename)}",
         ]
         table_args = connection.tables.map { |t| Shellwords.escape(t) }
@@ -80,16 +85,14 @@ module ActiveRecord
 
       private
 
-      def configuration
-        @configuration
-      end
+      attr_reader :configuration, :configuration_hash
 
       def default_collation
-        configuration["collation"] || DEFAULT_COLLATION
+        configuration_hash["collation"] || DEFAULT_COLLATION
       end
 
       def establish_master_connection
-        establish_connection configuration.merge("database" => "master")
+        establish_connection configuration_hash.merge("database" => "master")
       end
     end
 
@@ -110,9 +113,9 @@ module ActiveRecord
         end
 
         def configuration_host_ip(configuration)
-          return nil unless configuration["host"]
+          return nil unless configuration.host
 
-          Socket::getaddrinfo(configuration["host"], "echo", Socket::AF_INET)[0][3]
+          Socket::getaddrinfo(configuration.host, "echo", Socket::AF_INET)[0][3]
         end
 
         def local_ipaddr?(host_ip)


runephilosof-abtion:
Still does not seem to work 

This error is from the https://github.com/rails-sqlserver/ss_app, which I upgraded to Rails 6.1 https://github.com/runephilosof/ss_app/tree/rails6.1.

```
[vagrant@localhost ss_app]$ rails db:create
TinyTds::Error: Incorrect syntax near 'DATABASE'.
Couldn't create 'ss_app_development' database. Please check your configuration.
rails aborted!
ActiveRecord::StatementInvalid: TinyTds::Error: Incorrect syntax near 'DATABASE'.
/home/vagrant/ss_app/bin/rails:5:in `<top (required)>'
/home/vagrant/ss_app/bin/spring:10:in `block in <top (required)>'
/home/vagrant/ss_app/bin/spring:7:in `tap'
/home/vagrant/ss_app/bin/spring:7:in `<top (required)>'

Caused by:
TinyTds::Error: Incorrect syntax near 'DATABASE'.
/home/vagrant/ss_app/bin/rails:5:in `<top (required)>'
/home/vagrant/ss_app/bin/spring:10:in `block in <top (required)>'
/home/vagrant/ss_app/bin/spring:7:in `tap'
/home/vagrant/ss_app/bin/spring:7:in `<top (required)>'
Tasks: TOP => db:create
(See full trace by running task with --trace)
```

It also fails when running `rails db:drop`
```
[vagrant@localhost ss_app]$ rails db:drop
TinyTds::Error: Incorrect syntax near the keyword 'SET'.
Couldn't drop database 'ss_app_development'
rails aborted!
ActiveRecord::StatementInvalid: TinyTds::Error: Incorrect syntax near the keyword 'SET'.
/home/vagrant/ss_app/bin/rails:5:in `<top (required)>'
/home/vagrant/ss_app/bin/spring:10:in `block in <top (required)>'
/home/vagrant/ss_app/bin/spring:7:in `tap'
/home/vagrant/ss_app/bin/spring:7:in `<top (required)>'

Caused by:
TinyTds::Error: Incorrect syntax near the keyword 'SET'.
/home/vagrant/ss_app/bin/rails:5:in `<top (required)>'
/home/vagrant/ss_app/bin/spring:10:in `block in <top (required)>'
/home/vagrant/ss_app/bin/spring:7:in `tap'
/home/vagrant/ss_app/bin/spring:7:in `<top (required)>'
Tasks: TOP => db:drop:_unsafe
(See full trace by running task with --trace)
```

joshleblanc:
@runephilosof-abtion I believe you need to point your gemfile at the rails-6-1-support branch, it defaults to master.

Try something like `gem 'activerecord-sqlserver-adapter', github: "joshleblanc/activerecord-sqlserver-adapter", branch: "rails-6-1-support"`

runephilosof-abtion (approved):
@joshleblanc You are correct. How embarrassing. It is working perfectly.

runephilosof-abtion:
@wpolicarpo It would be nice to get this merged.
The tests are failing because of a pull rate limit with docker.com.
Rerun the tests?

Might want to authenticate to docker.com to increase the limit. But it should not delay this important PR.

vkalach:
@wpolicarpo any estimates when it will be merged?

wpolicarpo:
@joshleblanc can please you rebase your branch against `main`?

lib/active_record/connection_adapters/sqlserver/schema_statements.rb
```diff
@@ -422,10 +422,12 @@ def column_definitions(table_name)
               else
                 type = case ci[:type]
                        when /smallint|int|bigint/ then ci[:_type]
+                       when /bit/ then 'smallint'
```
wpolicarpo, line 425: Not sure if that was the best approach to solve that issue. The fact that you had to change a spec rings a bell that something might be wrong there.

TinyTDS casts the `bit` data type to boolean, can't see why we should change it at the adapter level. That could lead to unexpected bugs/changes to already existing apps assuming `boolean` as `true/false` instead of `1/0`.

In any case #875 addressed the original issue already.

lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -14,7 +14,7 @@ class SQLServerDatabaseTasks
                to: ActiveRecord::Base
 
       def initialize(configuration)
-        @configuration = configuration
+        @configuration = ActiveSupport::HashWithIndifferentAccess.new(configuration)
```
wpolicarpo, line 17: Why did you have to change it to a `HashWithIndifferentAccess`? Are you trying to mimic the new `using_database_configurations?` we should implement?

lib/active_record/connection_adapters/sqlserver/core_ext/array_handler.rb
```diff
@@ -0,0 +1,49 @@
...
+              end
+
+            unless nils.empty?
+              values_predicate = values_predicate.or(predicate_builder.build(attribute, nil))
+            end
+
+            array_predicates = ranges.map { |range| predicate_builder.build(attribute, range) }
+            array_predicates.unshift(values_predicate)
+            array_predicates.inject(&:or)
+          end
```
wpolicarpo, line 39: Can you explain the reasoning behind these changes? If possible, can you open a separate PR for each issue you are trying to address? It's easier for us to review and keep track of changes in future bug reports.

lib/active_record/tasks/sqlserver_database_tasks.rb
joshleblanc: After the 6.1 upgrade, none of the string keys were resolving. All of the keys were symbols. 

Changing it to a hash with indifferent access let me fix it with the least amount of disruptions.

lib/active_record/connection_adapters/sqlserver/core_ext/array_handler.rb
joshleblanc: This reverts an optimization made in rails 6.1, back to the functionality of 6.0. With the optimization present, any operations using an array would fail. 

For example, `where(id: [1,2,3])` would throw. 

I believe this was the commit that introduced the change: https://github.com/rails/rails/commit/72fd0bae5948c1169411941aeea6fef4c58f34a9

Sadly I don't remember the exact chain of events that lead me to this solution, it's been awhile.

lib/active_record/connection_adapters/sqlserver/schema_statements.rb
joshleblanc: This change came from #854

wpolicarpo: Right. #854 was closed via #875 and should not be introduced again.

wpolicarpo (changes requested):

joshleblanc:
Removed the 2 already-implemented fixes, updated the database tasks to use the configuration hash when required, rather than hash with indifferent access.

test/cases/column_test_sqlserver.rb
```diff
@@ -99,7 +99,7 @@ def assert_obj_set_and_save(attribute, value)
       _(col.sql_type).must_equal           "bit"
       _(col.type).must_equal               :boolean
       _(col.null).must_equal               true
-      _(col.default).must_equal            true
+      _(col.default).must_equal            1
```
wpolicarpo, line 102: Need to revert this one too.

PR #868, opened 2021-04-07
BigDecimal round change means test will no longer need coercing

aidanharan:
There was a change to BigDecimal#round (included in Ruby 3.0.0+) so that if argument < 1 then Integer is returned rather than a BigDecimal. With this change the 'MigrationTest#test_add_table_with_decimals' test will no longer need to be coerced in future.

See:
- https://github.com/ruby/bigdecimal/pull/170
- https://github.com/ruby/bigdecimal/blob/9067b353accb4dc407488460ccd1228e519d9185/CHANGES.md#202

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -396,37 +396,51 @@ module ActiveRecord
 end
 
 class MigrationTest < ActiveRecord::TestCase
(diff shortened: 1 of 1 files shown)


wpolicarpo:
Closed via #870.

Issue #859, opened 2021-03-19
Check if insert_all/upsert_all can be implemented using MERGE


wpolicarpo:
## Issue

Rails 6 introduced [`upsert_all`](https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/upsert_all) and the SQL Server adapter does not implement it yet.

There's no such thing as MySQL's `ON DUPLICATE KEY UPDATE` in SQL Server, but we could possibly use the [`MERGE`](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008/bb522522(v=sql.100)?redirectedfrom=MSDN) clause to achieve the same result.

## Expected behavior

`upsert_all` and `insert_all` could possibly be implemented using `MERGE`.

## Actual behavior

`upsert_all` and `insert_all` are not implemented.

mgrunberg:
@wpolicarpo I created PR https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/869 to explore this possibility. I found some issues.

andyundso:
Hi @aidanharan 

I would like to pick up the work by @mgrunberg  in #869, mostly because I would like to use Solid Queue with MSSQL. There is a recent comment by @justinko regarding a caveat.

> I think `MERGE` is still appropriate for this ... it just doesn't support duplicates in the "source". More info on that here: https://www.ibm.com/docs/en/informix-servers/14.10?topic=statement-handling-duplicate-rows

Question is if it would be okay if the `activerecord-sql-adapter` diverges from the other reference implementations (MySQL, SQLite and PostgreSQL) and de-duplicates the rows in the source (using the uniquely identifiable information, as suggested in the PR). I would assume most users never update the same thing multiple times within the same database query, but for those who do, it could be documented in the code that SQL server does this.

Otherwise, an official gem for the `insert` / `upsert` functionality also might be an idea to signal "hey, we can offer this functionality, but since it has a catch, it is its own thing".

aidanharan:
@andyundso Yes, I think de-duplicating the rows in the source in the adapter is fine if it can be made to work. We can document the difference to other adapters in the README. 

I'm no longer using MSSQL in work so I have limited time to give to the issue.

andyundso:
@aidanharan question about the tests: I do not really understand what this `coerce` does? does this disable tests from Rails upstream? how would I best go to add tests for this new `MERGE` mechanism?

[...]

Issue #847, opened 2020-10-19
ActiveRecord::ConnectionAdapters::SQLServerAdapter does not support skipping duplicates


Quintasan:
## Issue
`bundle exec rails db:seed` throws `ActiveRecord::ConnectionAdapters::SQLServerAdapter does not support skipping duplicates`.

## Expected behavior
A single SQL INSERT should be generated and executed

## Actual behavior
When using `#insert_all(Array[Hash])` I'm getting `ActiveRecord::ConnectionAdapters::SQLServerAdapter does not support skipping duplicates` 

## How to reproduce
```ruby

ENTRIES = [{name: "Test"}]
ENTRIES_DATA = ENTRIES.map { |e| e.merge(created_at: Time.zone.now, updated_at: Time.zone.now) }
Model.insert_all(ENTRIES_DATA)
```

## Details

- **Rails version**: `Rails 6.0.3.4`
- **SQL Server adapter version**: `6.0.0`
- **TinyTDS version**: `2.1.3`
- **FreeTDS details**: `1.1.6-1ubuntu0.1`

```
Compile-time settings (established with the "configure" script)
                            Version: freetds v1.1.6
             freetds.conf directory: /etc/freetds
     MS db-lib source compatibility: no
        Sybase binary compatibility: yes
                      Thread safety: yes
                      iconv library: yes
                        TDS version: auto
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: no
                             GnuTLS: yes
                               MARS: yes
```

jeffchuber:
@Quintasan did you find a solution or workaround for this? I'm seeing the same behavior with `PostgreSQLAdapter`. I worry it's an environment thing, because it works fine on my colleagues computer.

Quintasan:
The only workaround I could think of was using `find_or_create_by`.

jeffchuber:
I got "something" working... for some reason `insert_all!` works while `insert_all` does not. 🤷 🤦

mgrunberg:
Some info I gathered while working on the issue https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/859.

[...]

PR #869, opened 2021-04-07
insert_all/upsert_implementation using MERGE

mgrunberg:
This PR resolves the issue https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/859. Since it adds support to `insert_on_duplicate_skip` it also resolves https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/847.

The PR is in early stages. I'm exploring adding `supports_insert_on_duplicate_skip` first (insert_all). Upsert_all is still pending.

At this stage this test is falling with `ActiveRecord::RecordNotUnique: TinyTds::Error: Cannot insert duplicate key row in object 'dbo.books' with unique index 'index_books_on_author_id_and_name'. The duplicate key value is (8, Refactoring).
`
```ruby
def test_insert_all_with_skip_duplicates_and_autonumber_id_given
  skip unless supports_insert_on_duplicate_skip?

  assert_difference "Book.count", 1 do
    Book.insert_all [
      { id: 200, author_id: 8, name: "Refactoring" },
      { id: 201, author_id: 8, name: "Refactoring" }
    ]
  end
end
```

The test produces the following query
```sql
SET IDENTITY_INSERT [books] ON;
MERGE INTO [books] WITH (UPDLOCK, HOLDLOCK) AS target 
USING (SELECT DISTINCT * FROM (VALUES (200, 8, N'Refactoring'), (201, 8, N'Refactoring')) AS t1 ([id],[author_id],[name])) AS source 
ON (target.[author_id] = source.[author_id] AND target.[name] = source.[name]) OR (target.[id] = source.[id]) 
WHEN NOT MATCHED BY TARGET THEN 
  INSERT ([id],[author_id],[name]) VALUES (source.[id], source.[author_id], source.[name]) 
OUTPUT INSERTED.[id];
SET IDENTITY_INSERT [books] OFF;
```

SQL Server computes the source and target join and then applies the conditions to decide if a record from the joined table matches or not. In this case, both records are inserted.

I'm having doubts if implementing insert_all using merge is possible. 

upsert_all seems more challenging. Besides this problem, `WHEN MATCHED` can only update a row once.

For reference, table schema is
```ruby
create_table :books, id: :integer, force: true do |t|
    default_zero = { default: 0 }
    t.references :author
    t.string :format
    t.column :name, :string
    t.column :status, :integer, **default_zero
    t.column :read_status, :integer, **default_zero
    t.column :nullable_status, :integer
    t.column :language, :integer, **default_zero
    t.column :author_visibility, :integer, **default_zero
    t.column :illustrator_visibility, :integer, **default_zero
    t.column :font_size, :integer, **default_zero
    t.column :difficulty, :integer, **default_zero
    t.column :cover, :string, default: "hard"
    t.string :isbn, **case_sensitive_options
    t.datetime :published_on
    t.index [:author_id, :name], unique: true
    t.index :isbn, where: "published_on IS NOT NULL", unique: true
  end
```

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -140,6 +140,37 @@ module ActiveRecord
         private :default_insert_value
 
         def build_insert_sql(insert) # :nodoc:
+          if insert.skip_duplicates?
+            # Do we have a unique_by index? Use index columns
+            conflict_columns = if (unique_by = insert.send(:insert_all).unique_by)
+              [unique_by.columns]
+            else
+              # Compare against every unique constraint (primary key included).
+              # Discard constraints that are not fully included on insert.keys. Prevents invalid queries.
+              # Example: ignore unique index for columns ["name"] if insert keys is ["description"]
+              insert_all = insert.send(:insert_all)
+
+              (insert_all.send(:unique_indexes).map(&:columns) + [insert_all.primary_keys]).select do |columns|
+                columns.to_set.subset?(insert.keys)
+              end
+            end
+            includes_primary_key = (insert.send(:insert_all).primary_keys.to_set & insert.keys).present?
+
+            sql = +""
+            sql << "SET IDENTITY_INSERT #{insert.model.quoted_table_name} ON;" if includes_primary_key
+            sql << "MERGE INTO #{insert.model.quoted_table_name} WITH (UPDLOCK, HOLDLOCK) AS target"
+            sql << " USING (SELECT DISTINCT * FROM (#{insert.values_list}) AS t1 (#{insert.send(:columns_list)})) AS source"
+            sql << " ON (#{conflict_columns.map { |columns| columns.map { |column| "target.#{quote_column_name(column)} = source.#{quote_column_name(column)}" }.join(" AND ") }.join(") OR (")})"
+            sql << " WHEN NOT MATCHED BY TARGET THEN"
+            sql << " INSERT (#{insert.send(:columns_list)}) VALUES (#{insert.keys.map { |column| "source.#{quote_column_name(column)}" }.join(", ")})"
+            if returning = insert.send(:insert_all).returning
+              sql << " OUTPUT " << returning.map { |column| "INSERTED.#{quote_column_name(column)}" }.join(", ")
+            end
+            sql << ";"
+            sql << "SET IDENTITY_INSERT #{insert.model.quoted_table_name} OFF;" if includes_primary_key
+            return sql
+          end
+
           sql = +"INSERT #{insert.into}"
 
           if returning = insert.send(:insert_all).returning
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -169,7 +169,7 @@ module ActiveRecord
       end
 
       def supports_insert_on_duplicate_skip?
-        false
+        true
       end
 
       def supports_insert_on_duplicate_update?
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1555,3 +1555,8 @@ class ReloadModelsTest < ActiveRecord::TestCase
   # `activesupport/lib/active_support/testing/isolation.rb` exceeds what Windows can handle.
   coerce_tests! :test_has_one_with_reload if RbConfig::CONFIG["host_os"] =~ /mswin|mingw/
 end
+
+class InsertAllTest < ActiveRecord::TestCase
+  # Skip this until upsert is supported
+  coerce_tests! :test_insert
+end
--- a/test/cases/insert_all_test_sqlserver.rb
+++ b/test/cases/insert_all_test_sqlserver.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require "cases/helper_sqlserver"
+require "models/book"
+
+class InsertAllTestSQLServer < ActiveRecord::TestCase
+  fixtures :books
+
+  # Issue https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/847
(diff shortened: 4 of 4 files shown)


lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -140,6 +140,37 @@ def default_insert_value(column)
...
+
+              (insert_all.send(:unique_indexes).map(&:columns) + [insert_all.primary_keys]).select do |columns|
+                columns.to_set.subset?(insert.keys)
+              end
+            end
+            includes_primary_key = (insert.send(:insert_all).primary_keys.to_set & insert.keys).present?
+
+            sql = +""
+            sql << "SET IDENTITY_INSERT #{insert.model.quoted_table_name} ON;" if includes_primary_key
+            sql << "MERGE INTO #{insert.model.quoted_table_name} WITH (UPDLOCK, HOLDLOCK) AS target"
```
gisborne, line 161: I believe HOLDLOCK is equivalent to SERIALIZABLE. I'd prefer the SQL standard term. Unless HOLDLOCK is usual in the SQL Server community?

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -140,6 +140,37 @@ def default_insert_value(column)
         private :default_insert_value
 
         def build_insert_sql(insert) # :nodoc:
+          if insert.skip_duplicates?
+            # Do we have a unique_by index? Use index columns
+            conflict_columns = if (unique_by = insert.send(:insert_all).unique_by)
```
gisborne, line 145: Can we not do `insert.insert_all`?

gisborne: Suggest we cache insert_all in a local variable rather than have to look it up each time.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -140,6 +140,37 @@ def default_insert_value(column)
...
+              (insert_all.send(:unique_indexes).map(&:columns) + [insert_all.primary_keys]).select do |columns|
+                columns.to_set.subset?(insert.keys)
+              end
+            end
+            includes_primary_key = (insert.send(:insert_all).primary_keys.to_set & insert.keys).present?
+
+            sql = +""
+            sql << "SET IDENTITY_INSERT #{insert.model.quoted_table_name} ON;" if includes_primary_key
+            sql << "MERGE INTO #{insert.model.quoted_table_name} WITH (UPDLOCK, HOLDLOCK) AS target"
+            sql << " USING (SELECT DISTINCT * FROM (#{insert.values_list}) AS t1 (#{insert.send(:columns_list)})) AS source"
```
gisborne, line 162: None of the standard adapters uniquify insert rows. I don't think we need to do it here either.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -140,6 +140,37 @@ def default_insert_value(column)
...
+            sql << " ON (#{conflict_columns.map { |columns| columns.map { |column| "target.#{quote_column_name(column)} = source.#{quote_column_name(column)}" }.join(" AND ") }.join(") OR (")})"
+            sql << " WHEN NOT MATCHED BY TARGET THEN"
+            sql << " INSERT (#{insert.send(:columns_list)}) VALUES (#{insert.keys.map { |column| "source.#{quote_column_name(column)}" }.join(", ")})"
+            if returning = insert.send(:insert_all).returning
+              sql << " OUTPUT " << returning.map { |column| "INSERTED.#{quote_column_name(column)}" }.join(", ")
+            end
+            sql << ";"
+            sql << "SET IDENTITY_INSERT #{insert.model.quoted_table_name} OFF;" if includes_primary_key
+            return sql
```
gisborne, line 171: Repeatedly appending to a string is somewhat slow. Faster in that case is to append strings to an array and then join the array at the end. Not sure how much difference that would make, but it would be just as easy to read, I think, and would probably be noticeable if we are dealing with many rows.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
mgrunberg: I will check that. Thanks!

lib/active_record/connection_adapters/sqlserver/database_statements.rb
mgrunberg: insert_all is private, it's not possible to do `insert.insert_all`. 
Thanks for the suggestion but I'm still not paying attention to things like this. As I said, the PR is in an early stage. I have plans to improve the code (refactor) but right now I'm focused on produce the SQL that cover all tests.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
mgrunberg: uniquify was the solution to this test/scenario
```
Book.insert_all [
  { author_id: 8, name: "Refactoring" },
  { author_id: 8, name: "Refactoring" }
]
```
I don't like it but it works so I move on until all test pass.

Then I found the failing test (still stuck with it)
```
Book.insert_all [
  { id: 200, author_id: 8, name: "Refactoring" },
  { id: 201, author_id: 8, name: "Refactoring" }
]
```
My impression is that the solution to this will let me remove the uniquify.

justinko:
I think `MERGE` is still appropriate for this ... it just doesn't support duplicates in the "source".  More info on that here: https://www.ibm.com/docs/en/informix-servers/14.10?topic=statement-handling-duplicate-rows

Would a gem (e.g. `active record-sqlserver-adapter-insert-all`) make sense as a replacement for this pull? There would of course be a caveat stated that your source/args for insertion cannot contain duplicates.

mgrunberg:
The feature was implemented in #1312

PR #878, opened 2021-04-14
Add some more niceties to the database tasks

bf4:


lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -70,11 +70,14 @@ def structure_dump(filename, extra_flags)
         dump.gsub!(/^GO\n/, "")                               # Strip db GO statements
         dump.gsub!(/nvarchar\(8000\)/, "nvarchar(4000)")      # Fix nvarchar(8000) column defs
         dump.gsub!(/nvarchar\(-1\)/, "nvarchar(max)")         # Fix nvarchar(-1) column defs
+        dump.gsub!(/varbinary\(-1\)/, "varbinary(max)")       # Fix varbinary(-1) column defs
         dump.gsub!(/text\(\d+\)/, "text")                     # Fix text(16) column defs
+        dump.gsub!(/\r\n/, "\n")                              # Convert line breaks
+        dump.gsub!(/\s+$/, "")                                # Strip trailing whitespace
```
bf4, line 76: I'm using these two.. probably don't need to opt-everyone in to them, but thought I'd share

lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -70,11 +70,14 @@ def structure_dump(filename, extra_flags)
         dump.gsub!(/^GO\n/, "")                               # Strip db GO statements
         dump.gsub!(/nvarchar\(8000\)/, "nvarchar(4000)")      # Fix nvarchar(8000) column defs
         dump.gsub!(/nvarchar\(-1\)/, "nvarchar(max)")         # Fix nvarchar(-1) column defs
+        dump.gsub!(/varbinary\(-1\)/, "varbinary(max)")       # Fix varbinary(-1) column defs
```
bf4, line 73: 👀 came here for this

lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -59,9 +59,9 @@ def structure_dump(filename, extra_flags)
           "-P #{Shellwords.escape(configuration['password'])}",
           "-o #{Shellwords.escape(filename)}",
         ]
-        table_args = connection.tables.map { |t| Shellwords.escape(t) }
+        table_args = connection.tables.sort.map { |t| Shellwords.escape(t) }
```
bf4, line 62: figure it makes it more deterministic

lib/active_record/tasks/sqlserver_database_tasks.rb
aidanharan: Does this fix a bug you were having with a specific version of Rails (6.0/6.1/etc)? 

Would you be able to update `test/cases/rake_test_sqlserver.rb` to demonstrate the fix?

bf4: yes, wil do.

bf4: Not a Rails version issue. Just some database we're querying on.

lib/active_record/tasks/sqlserver_database_tasks.rb
```diff
@@ -59,9 +59,9 @@ def structure_dump(filename, extra_flags)
...
         command.concat(table_args)
-        view_args = connection.views.map { |v| Shellwords.escape(v) }
+        view_args = connection.views.sort.map { |v| Shellwords.escape(v) }
```
wpolicarpo, line 64: This one is tricky. What if you have a view that depends on another view but alphabetical order does not match? Wouldn't that be a problem?

lib/active_record/tasks/sqlserver_database_tasks.rb
wpolicarpo: Can you maybe add that test case?

bf4:
I intend to get to it.  Thanks for checking in.

PR #879, opened 2021-04-15
Added visit method for HomogeneousIn Arel type

aidanharan:
Added visit method for the new Arel node type `HomogenousIn` used when the values are all homogenous (see https://github.com/rails/rails/pull/39009).

This change means we don't need to fallback to an old version of `ArrayHandler` as proposed in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/861

Also, cleaned up some method signatures.

**Before**
https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2343775735
7133 runs, 16254 assertions, 37 failures, 805 errors, 37 skips

**After**
https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/879/checks?check_run_id=2351710950
7133 runs, 18457 assertions, 25 failures, 145 errors, 37 skips

### DIFF
--- a/lib/arel/visitors/sqlserver.rb
+++ b/lib/arel/visitors/sqlserver.rb
@@ -18,10 +18,6 @@ module Arel
 
       def bind_block; BIND_BLOCK; end
 
-      def visit_Arel_Nodes_BindParam(o, collector)
-        collector.add_bind(o.value, &bind_block)
-      end
-
       def visit_Arel_Nodes_Bin(o, collector)
         visit o.expr, collector
         collector << " #{ActiveRecord::ConnectionAdapters::SQLServerAdapter.cs_equality_operator} "
--- a/lib/arel/visitors/sqlserver.rb
+++ b/lib/arel/visitors/sqlserver.rb
@@ -11,13 +11,18 @@ module Arel
 
       private
 
-      # SQLServer ToSql/Visitor (Overides)
+      # SQLServer ToSql/Visitor (Overrides)
 
-      def visit_Arel_Nodes_BindParam o, collector
-        collector.add_bind(o.value) { |i| "@#{i - 1}" }
+      BIND_BLOCK = proc { |i| "@#{i - 1}" }
+      private_constant :BIND_BLOCK
+
+      def bind_block; BIND_BLOCK; end
+
+      def visit_Arel_Nodes_BindParam(o, collector)
+        collector.add_bind(o.value, &bind_block)
       end
 
-      def visit_Arel_Nodes_Bin o, collector
+      def visit_Arel_Nodes_Bin(o, collector)
         visit o.expr, collector
         collector << " #{ActiveRecord::ConnectionAdapters::SQLServerAdapter.cs_equality_operator} "
       end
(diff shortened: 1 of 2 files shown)


lib/arel/visitors/sqlserver.rb
```diff
@@ -11,13 +11,18 @@ class SQLServer < Arel::Visitors::ToSql
...
 
-      # SQLServer ToSql/Visitor (Overides)
+      # SQLServer ToSql/Visitor (Overrides)
 
-      def visit_Arel_Nodes_BindParam o, collector
-        collector.add_bind(o.value) { |i| "@#{i - 1}" }
+      BIND_BLOCK = proc { |i| "@#{i - 1}" }
+      private_constant :BIND_BLOCK
+
+      def bind_block; BIND_BLOCK; end
```
wpolicarpo, line 19: Do we really have to redefine `bind_block` method?

lib/arel/visitors/sqlserver.rb
```diff
@@ -11,13 +11,18 @@ class SQLServer < Arel::Visitors::ToSql
...
+
+      def visit_Arel_Nodes_BindParam(o, collector)
```
wpolicarpo, line 21: With `bind_block` introduction, we can remove `visit_Arel_Nodes_BindParam` now that it has the same definition as the parent Abstract adapter.

lib/arel/visitors/sqlserver.rb
```diff
@@ -63,7 +68,36 @@ def visit_Arel_Nodes_Grouping(o, collector)
...
+        else
+          collector << " NOT IN ("
+        end
+
+        values = o.casted_values
+
+        if values.empty?
+          collector << @connection.quote(nil)
+        else
+          # Monkey-patch start. Add query attribute bindings rather than just values.
```
wpolicarpo, line 87: Does that mean that Rails 6.1 is not parameterizing queries with `IN` clauses anymore?

wpolicarpo (changes requested):

lib/arel/visitors/sqlserver.rb
aidanharan: Yes, you need to redefine it unfortunately.

```ruby
class Car
  COLOUR = 'white'

  def colour
    COLOUR
  end
end

class Ferrari < Car
  COLOUR = 'red'
end
```

```
Car.new.colour
=> white
puts Ferrari.new.colour
=> white
```

lib/arel/visitors/sqlserver.rb
aidanharan: Rails 6.1 is still parameterizing queries with IN clauses. The difference between PostgreSQL/MySQL/SQLite and SQL Server is that we require the type information for the query.

```ruby
Topic.where(id: [1, 2])
```

PostgreSQL
> SELECT "topics".* FROM "topics" WHERE "topics"."id" IN ($1, $2)

MySQL
> SELECT `topics`.* FROM `topics` WHERE `topics`.`id` IN (1, 2)

SQLite
> SELECT "topics".* FROM "topics" WHERE "topics"."id" IN (?, ?)

SQL Server
> EXEC sp_executesql N'SELECT [topics].* FROM [topics] WHERE [topics].[id] IN (@0, @1)', N'@0 bigint, @1 bigint', @0 = 1, @1 = 2

lib/arel/visitors/sqlserver.rb
aidanharan: Removed.

Issue #848, opened 2020-11-24
Column Collation not being set in db:schema:load


pfriedman:
## Issue

When attempting to set the collation for a column[1] that differs from the collation for the schema, the column collation gets set correctly when running `rake db:migrate` but running `rake db:schema:load` the column collation reverts back to the default schema location.

[1] https://docs.microsoft.com/en-us/sql/relational-databases/collations/set-or-change-the-column-collation

## Expected behavior

With a migration file that attempts to set the collation for a particular column, e.g. 

```
  change_column :models, :column_name, 'VARCHAR(256) COLLATE SQL_Latin1_General_CP1_CS_AS'
```

and the default collation being different (SQL_Latin1_General_CP1_CI_AS) (CI not CS) 
we expect that the collation for the column is the one we set in the change_column method

```
> Model.columns.select { |a| a.name == "column_name" }
=> [#<ActiveRecord::ConnectionAdapters::SQLServerColumn:0x0000...
  @collation="SQL_Latin1_General_CP1_CS_AS",
  @name="column_name",
```

ALSO in the `db/schema.rb` file we see that the collation comes across as a constant where it really should be a quoted String

```
t.varchar "column_name", limit: 256, collation: SQL_Latin1_General_CP1_CS_AS
```

where we'd expect it to look more like 

```
t.varchar "column_name", limit: 256, collation: "SQL_Latin1_General_CP1_CS_AS"
```

## Actual behavior
When running `rake db:schema:load`, the collation is set to the default 

```
> Model.columns.select { |a| a.name == "column_name" }
=> [#<ActiveRecord::ConnectionAdapters::SQLServerColumn:0x0000...
  @collation="SQL_Latin1_General_CP1_CI_AS",
  @name="column_name",
```

## Current Patch 
In our project we have the following change in order to "fix" this issue 

```
require 'activerecord-sqlserver-adapter'

# Constant set to bypass UninitializedConstant error
SQL_Latin1_General_CP1_CS_AS = 'SQL_Latin1_General_CP1_CS_AS'

module ActiveRecord
  module ConnectionAdapters
    module SQLServer
      class SchemaCreation
        def add_column_options!(sql, options)
          sql << " DEFAULT #{quote_default_expression(options[:default], options[:column])}" if options_include_default?(options)
          if options[:null] == false
            sql << " NOT NULL"
          end
          if options[:collation].present?
            sql << " COLLATE #{options[:collation]}"
          end
[...]

PR #881, opened 2021-04-16
Dump column collation to schema.rb and allow collation changes using column_change

mgrunberg:
This PR address https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/848. 

It also adds support to change collation `change_column :customers, :name, collation: :xxx`. Before this change, collation changes were skipped/ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [#879](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/879) Added visit method for HomogeneousIn
 - [#880](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/880) Handle any default column class when deduplicating
 - [#861](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/861) Fix Rails 6.1 database config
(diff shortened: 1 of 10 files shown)


mgrunberg:
I decided to implement this on 6-0-stable since it's a small change. If it wasn't the right call and it should be targeting main, let me know.

mgrunberg:
I've just realized that the issue was reported for 5-2-stable. 

The fix is almost the same. The only difference with this PR is that the change in `lib/active_record/connection_adapters/sqlserver/schema_statements.rb` should be
```
sql_commands.last << " COLLATE #{options[:collation]}" if options[:collation].present?
```
and not
```
alter_command += " COLLATE #{options[:collation]}" if options[:collation].present?
```

I'm not sure what's the best here. I can make another PR for 5-2-stable.

wpolicarpo:
PRs should target the `main` branch. We can backport to stable branches if necessary. Can you rebase your branch and change the target to `main`?

mgrunberg:
@wpolicarpo, no problem. I've rebase with main and change target to main.

wpolicarpo:
LGTM. Can you resolve the changelog conflict?

mgrunberg:
@wpolicarpo, conflicts fixed

PR #882, opened 2021-04-17
Rails 6.1 fixes: Use renamed connection_config method and rewrite coerced tests using unsafe raw SQL

mgrunberg:
This PR contains two fixes:

1. ARTest.connfiguration_config was renamed (see https://github.com/rails/rails/pull/38007)
2. Rewrite `UnsafeRawSqlTest` coerced tests because unsafe raw SQL support was removed (see https://github.com/rails/rails/commit/317465a4a8f8c0ebaf9534921350c4540584c03c)


wpolicarpo:
Thanks for the contribution.

PR #901, opened 2021-04-21
Rails 6.1: coerce tests when we produce more or less queries

mgrunberg:
PR fixes the following failures:

```
BasePreventWritesTest::BasePreventWritesLegacyTest#test_0006_an empty transaction does not raise if preventing writes [/usr/local/bundle/bundler/gems/rails-8b63ea762239/activerecord/test/cases/base_prevent_writes_test.rb:164]:
1 instead of 2 queries were executed.
```
```
BasePreventWritesTest#test_0006_an empty transaction does not raise if preventing writes [/usr/local/bundle/bundler/gems/rails-8b63ea762239/activerecord/test/cases/base_prevent_writes_test.rb:61]:
1 instead of 2 queries were executed.
```
```
ActiveRecord::AdapterPreventWritesLegacyTest#test_errors_when_an_insert_query_is_called_while_preventing_writes [/usr/local/bundle/bundler/gems/rails-8b63ea762239/activerecord/test/cases/adapter_prevent_writes_test.rb:181]:
6 instead of 0 queries were executed.
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -47,10 +47,17 @@ end
 module ActiveRecord
   class AdapterPreventWritesTest < ActiveRecord::TestCase
     # Fix randomly failing test. The loading of the model's schema was affecting the test.
+    # Also, we do some read queries. Remove assert_no_queries
     coerce_tests! :test_errors_when_an_insert_query_is_called_while_preventing_writes
     def test_errors_when_an_insert_query_is_called_while_preventing_writes_coerced
       Subscriber.send(:load_schema!)
-      original_test_errors_when_an_insert_query_is_called_while_preventing_writes
+      assert_raises(ActiveRecord::ReadOnlyError) do
+        ActiveRecord::Base.while_preventing_writes do
+          @connection.transaction do
+            @connection.insert("INSERT INTO subscribers(nick) VALUES ('138853948594')", nil, false)
+          end
+        end
+      end
     end
   end
 end
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1674,3 +1674,31 @@ class MarshalSerializationTest < ActiveRecord::TestCase
     )
   end
 end
+
+class BasePreventWritesTest < ActiveRecord::TestCase
+  # We open one transaction, not two. Same as original but checking one query
+  coerce_tests! %r{an empty transaction does not raise if preventing writes}
+  test "an empty transaction does not raise if preventing writes coerced" do
+    ActiveRecord::Base.while_preventing_writes do
+      assert_queries(1, ignore_none: true) do
+        Bird.transaction do
+          ActiveRecord::Base.connection.materialize_transactions
+        end
+      end
+    end
+  end
+end
+
+class BasePreventWritesLegacyTest < ActiveRecord::TestCase
+  # We open one transaction, not two. Same as original but checking one query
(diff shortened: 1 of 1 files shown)


test/cases/coerced_tests.rb
```diff
@@ -47,10 +47,17 @@ def test_value_limit_violations_are_translated_to_specific_exception_coerced
 module ActiveRecord
   class AdapterPreventWritesTest < ActiveRecord::TestCase
     # Fix randomly failing test. The loading of the model's schema was affecting the test.
+    # Also, we do some read queries. Remove assert_no_queries
```
wpolicarpo, line 50: What kind read queries we do? The fact that we are coercing tests like this make me fell we are missing some `SCHEMA` tags when querying for internal information.

test/cases/coerced_tests.rb
```diff
@@ -1667,3 +1674,31 @@ def marshal_fixture_path(file_name)
     )
   end
 end
+
+class BasePreventWritesTest < ActiveRecord::TestCase
+  # We open one transaction, not two. Same as original but checking one query
```
wpolicarpo, line 1679: Do you know why there's such behaviour difference?

test/cases/coerced_tests.rb
```diff
@@ -1667,3 +1674,31 @@ def marshal_fixture_path(file_name)
...
+        Bird.transaction do
+          ActiveRecord::Base.connection.materialize_transactions
+        end
+      end
+    end
+  end
+end
+
+class BasePreventWritesLegacyTest < ActiveRecord::TestCase
+  # We open one transaction, not two. Same as original but checking one query
```
wpolicarpo, line 1693: Same here. Do you know why we open only 1 transaction?

test/cases/coerced_tests.rb
aidanharan: I don't see this test failing in the latest CI build (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2399437373). The issue with unexpected read queries was fixed in the past by `Subscriber.send(:load_schema!)` so I don't know why it would be no longer working. 

I think the changes to `test_errors_when_an_insert_query_is_called_while_preventing_writes_coerced` should be removed and if the issue happens again on the CI we can look at it again.

test/cases/coerced_tests.rb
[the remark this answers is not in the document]
mgrunberg: Thanks for helping with the explanation. Sure, I changed the comment

test/cases/coerced_tests.rb
mgrunberg: Mmmm, but I do see the test failing in some runs https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2400618957?check_suite_focus=true#step:4:440 (my last commit run).

I checked the last run and you were right, it wasn't failing. But there was another one failing https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2399437340?check_suite_focus=true#step:4:695. I coerced that test following your suggestion about using `Subscriber.send(:load_schema!)`.

In some ruby versions the test is still failing https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2400618873?check_suite_focus=true#step:4:448

And in other ruby version the original (the one I changed) is failing https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/2400618957?check_suite_focus=true#step:4:440.

I can't explain this.

PR #902, opened 2021-04-21
Rails 6.1: change schema dumper condition to only dump explicit default for non identity integers

mgrunberg:
PR fixes:

```
PrimaryKeyAnyTypeTest#test_0001_schema dump primary key includes type and options [/usr/local/bundle/bundler/gems/rails-8b63ea762239/activerecord/test/cases/primary_keys_test.rb:315]:
Expected /create_table "barcodes", primary_key: "code", id: { type: :string, limit: 42 }/ to match "....ActiveRecord::Schema.define(version: 0) do\n\n  create_table \"barcodes\", primary_key: \"code\", id: { type: :string, limit: 42, default: nil }, force: :cascade do |t|\n  end\n\nend\n"
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1509,6 +1509,25 @@ class RelationMergingTest < ActiveRecord::TestCase
     relation = Post.all.merge(Post.order([Arel.sql("title LIKE ?"), "%suffix"]))
     assert_equal ["title LIKE N'%suffix'"], relation.order_values
   end
+
+  # Same as original but change first regexp to match sp_executesql binding syntax
+  coerce_tests! :test_merge_doesnt_duplicate_same_clauses
+  def test_merge_doesnt_duplicate_same_clauses_coerced
+    david, mary, bob = authors(:david, :mary, :bob)
+
+    non_mary_and_bob = Author.where.not(id: [mary, bob])
+
+    author_id = Author.connection.quote_table_name("authors.id")
+    assert_sql(/WHERE #{Regexp.escape(author_id)} NOT IN \((@\d), \g<1>\)'/) do
+      assert_equal [david], non_mary_and_bob.merge(non_mary_and_bob)
+    end
+
+    only_david = Author.where("#{author_id} IN (?)", david)
+
+    assert_sql(/WHERE \(#{Regexp.escape(author_id)} IN \(1\)\)\z/) do
+      assert_equal [david], only_david.merge(only_david)
+    end
+  end
 end
 
 module ActiveRecord
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1509,25 +1509,6 @@ class RelationMergingTest < ActiveRecord::TestCase
     relation = Post.all.merge(Post.order([Arel.sql("title LIKE ?"), "%suffix"]))
     assert_equal ["title LIKE N'%suffix'"], relation.order_values
   end
-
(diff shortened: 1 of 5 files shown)


test/cases/coerced_tests.rb
```diff
@@ -1667,3 +1686,13 @@ def marshal_fixture_path(file_name)
     )
   end
 end
+
+class PrimaryKeyAnyTypeTest < ActiveRecord::TestCase
+  # Same as original but add 'default: nil' option
+  coerce_tests! %r{schema dump primary key includes type and options}
+  test "schema dump primary key includes type and options coerced" do
+    schema = dump_table_schema "barcodes"
+    assert_match %r/create_table "barcodes", primary_key: "code", id: { type: :string, limit: 42, default: nil }/, schema
```
wpolicarpo, line 1695: Do you know why we differ from the original test?

mgrunberg: The difference is due to 
https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver/schema_dumper.rb#L18.

That method was introduced by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/commit/89ccbe40c611128d9524a11054001eb431c2c1f0.

Now I'm thinking that maybe instead of coercing we should change that condition. I'm not sure what's the best here.

Here are the definitions of other adapters:

[MySQL](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/mysql/schema_dumper.rb#L37)
```
column.type == :integer && !column.auto_increment?
```
[Postgresql](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb)
```
column.type == :uuid || (column.type == :integer && !column.serial?)
```

[Sqlite3](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/sqlite3/schema_dumper.rb)
```
column.bigint?
```

By checking `primary_key?` and not ` column.type` string primary keys end up having the explicit default.

By the way, I split this PR into two.

mgrunberg: I did that. Also made small changes not related to the schema difference but related to primary_key handling:

1. Remove one condition from `default_primary_key?` (redundant from my point of view). 
2. Change `primary_key_non_clustered` type to bigint
3. Remove unnecessary option set to `uuid` primary_keys

Sorry if they were out of scope.

I added a test similar to [the one rails](https://github.com/rails/rails/blob/v6.1.0/activerecord/test/cases/primary_keys_test.rb#L419) has to show that:
1. bigint and integer primary_keys have `:integer` type (I trust on that in `explicit_primary_key_default?`)
2. bigint identity column is the default primary key and not dumped to schema.

test/cases/primary_keys_test_sqlserver.rb
```diff
@@ -0,0 +1,103 @@
...
+    assert_predicate column, :is_primary?
+    assert_predicate column, :is_identity?
+    assert_equal :integer, column.type
+    assert_predicate column, :bigint?
+
+    schema = dump_table_schema "widgets"
+    assert_match %r/create_table "widgets", force: :cascade do/, schema
+  end
+
+  test "don't set itentity to integer and bigint when there is a default" do
```
wpolicarpo, line 85: Typo `itentity`.

wpolicarpo (commented):
Can you also rebase your branch?

PR #903, opened 2021-04-21
Rails 6.1: Raise `ActiveRecord::ConnectionNotEstablished` on calls to execute with a disconnected connection.

mgrunberg:
This PR fixes the failure:
```
TestDisconnectedAdapter#test_0001_can't execute statements while disconnected [/usr/local/bundle/bundler/gems/rails-8b63ea762239/activerecord/test/cases/disconnected_test.rb:25]:
[ActiveRecord::ConnectionNotEstablished] exception expected, not
Class: <ActiveRecord::StatementInvalid>
Message: <"NoMethodError: undefined method `execute' for nil:NilClass\nDid you mean?  expects">
```
Rails 6.1 expects that we raise an instance of `ActiveRecord::ConnectionNotEstablished` (see https://github.com/rails/rails/blob/v6.1.0/activerecord/test/cases/disconnected_test.rb).
Rails 6.0 was expecting `ActiveRecord::StatementInvalid`. In fact, 6-0-stable raises that exception with the message: 
```
"NoMethodError: undefined method `execute' for nil:NilClass\nDid you mean?  expects" 
```

I use the fact that @connnection is nil to detect when to raise the exception.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -464,7 +464,7 @@ module ActiveRecord
         end
 
         def ensure_established_connection!
-          raise TinyTds::Error, 'SQL Server client is not connected' if @connection.nil? || !@connection.active?
+          raise TinyTds::Error, 'SQL Server client is not connected' unless @connection
 
           yield
         end
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -464,7 +464,7 @@ module ActiveRecord
         end
 
         def ensure_established_connection!
-          raise ActiveRecord::ConnectionNotEstablished, 'SQL Server client is not connected' unless @connection
+          raise ActiveRecord::ConnectionNotEstablished, 'SQL Server client is not connected' if @connection.nil? || !@connection.active?
 
           yield
         end
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -167,7 +167,7 @@ module ActiveRecord
           log(sql, name) do
             case @connection_options[:mode]
             when :dblib
-              result = ensure_established_connection! { @connection.execute(sql) }
+              result = ensure_established_connection! { dblib_execute(sql) }
               options = { as: :hash, cache_rows: true, timezone: ActiveRecord::Base.default_timezone || :utc }
               result.each(options) do |row|
                 r = row.with_indifferent_access
@@ -357,13 +357,7 @@ module ActiveRecord
         def raw_connection_do(sql)
           case @connection_options[:mode]
           when :dblib
-            result = ensure_established_connection! { @connection.execute(sql) }
-
-            # TinyTDS returns false instead of raising an exception if connection fails.
-            # Getting around this by raising an exception ourselves while this PR
-            # https://github.com/rails-sqlserver/tiny_tds/pull/469 is not released.
-            raise TinyTds::Error, "failed to execute statement" if result.is_a?(FalseClass)
-
+            result = ensure_established_connection! { dblib_execute(sql) }
             result.do
           end
         ensure
@@ -428,7 +422,7 @@ module ActiveRecord
         def raw_connection_run(sql)
           case @connection_options[:mode]
           when :dblib
-            ensure_established_connection! { @connection.execute(sql) }
+            ensure_established_connection! { dblib_execute(sql) }
           end
         end
 
@@ -461,6 +455,15 @@ module ActiveRecord
             handle.cancel if handle
           end
           handle
+        end
+
+        def dblib_execute(sql)
+          @connection.execute(sql).tap do |result|
+            # TinyTDS returns false instead of raising an exception if connection fails.
+            # Getting around this by raising an exception ourselves while this PR
+            # https://github.com/rails-sqlserver/tiny_tds/pull/469 is not released.
+            raise TinyTds::Error, "failed to execute statement" if result.is_a?(FalseClass)
+          end
         end
 
         def ensure_established_connection!
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -464,7 +464,7 @@ module ActiveRecord
         end
 
         def ensure_established_connection!
-          raise ActiveRecord::ConnectionNotEstablished, 'SQL Server client is not connected' if @connection.nil? || !@connection.active?
+          raise TinyTds::Error, 'SQL Server client is not connected' if @connection.nil? || !@connection.active?
 
           yield
         end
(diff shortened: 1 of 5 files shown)


lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -167,6 +167,8 @@ def execute_procedure(proc_name, *variables)
           log(sql, name) do
             case @connection_options[:mode]
             when :dblib
+              raise ActiveRecord::ConnectionNotEstablished if @connection.nil?
```
wpolicarpo, line 170: I need to investigate it further, but I'm not sure if checking for `@connection.nil?` is enough. What happens if the `@connection` is not `nil` but connection with socket is dead? Looks like this is the way they found to fix issues like https://github.com/rails-sqlserver/tiny_tds/pull/469.

Also, I see they've introduced a `.new_client` method encapsulating the underlying connection instantiation and I think it would be good to implement the same pattern here.

test/cases/disconnected_test_sqlserver.rb
```diff
@@ -0,0 +1,39 @@
...
+    @connection = ActiveRecord::Base.connection
+  end
+
+  teardown do
+    return if in_memory_db?
+    db_config = ActiveRecord::Base.connection_db_config
+    ActiveRecord::Base.establish_connection(db_config)
+  end
+
+  test "can't execute procuderes while disconnected" do
```
gisborne, line 18: Typo

lib/active_record/connection_adapters/sqlserver/database_statements.rb
mgrunberg: You are right, `@connection.nil?` only deals with calls to closed connections ([we set it to nil](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver_adapter.rb#L214)). I added `connection.active?` check to deal with dead/closed connections. I guess that if those conditions are ok and later the statement fail (tiny_tds returns false) it's fair to raise something different than `ActiveRecord::ConnectionNotEstablished`. What do you think?

I will look the `.new_client` pattern and implement it.

aidanharan: In my opinion:

1) We should not check if a connection is dead/alive before using it. Using `connection.active?` would mean an extra call to the database for every query. We will know whether the connection is dead based on whether `TinyTds::Error("failed to execute statement")` is thrown. So I would remove `!@connection.active?` from https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/903/files#diff-f749b206a3bbfaa7e02aa77f1c26715bea21134aa6d2645cbf9068783e280cf8R467

2) If a connection is disconnected or dead then a `ConnectionNotEstablished` should be thrown. The disconnection part is already done. We just need to check if the connection was dead. This can be detected by checking for the "failed to execute statement" error message when translating the exception (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/903/files#diff-3562da8bff63a7b06dce2a8b9e1163ab5598cca3bcaaf1b971a1c3e40942b053R378). 

3) The refactoring of the connection instantiation to use the `.new_client` pattern should be opened as a separate issue.

aidanharan: I've included my suggestions as PR https://github.com/yellowspot/activerecord-sqlserver-adapter/pull/2

mgrunberg: @aidanharan 
> In my opinion:
> 
> 1. We should not check if a connection is dead/alive before using it. Using `connection.active?` would mean an extra call to the database for every query. We will know whether the connection is dead based on whether `TinyTds::Error("failed to execute statement")` is thrown. So I would remove `!@connection.active?` from https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/903/files#diff-f749b206a3bbfaa7e02aa77f1c26715bea21134aa6d2645cbf9068783e280cf8R467

Are you sure that checking `connection.active?` means an extra call to the database? I thought that method was doing a light check using the client's status (no roundtrip to the database at all). 

That being said, I see a valid point in not checking dead connections and let them raise errors. The other main adapters are doing that. 

I also thought about doing that. My concern was that detecting "failed to execute statement" as `ConnectionNotEstablished` could be too general (be inappropriate sometimes). I mean, other adapters are detecting more specific situations: [MySQL client is not connected](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L639), [connection is closed](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L477) 
(postgres) and [called on a closed database](https://github.com/rails/rails/blob/v6.1.0/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L478) (sqlite3).

What do you think about this? 

> 2. If a connection is disconnected or dead then a `ConnectionNotEstablished` should be thrown. The disconnected part is already done. We just need to check if the connection is dead. This can be detected by checking for the "failed to execute statement" error message when translating the exception (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/903/files#diff-3562da8bff63a7b06dce2a8b9e1163ab5598cca3bcaaf1b971a1c3e40942b053R378).
> 3. The refactoring of the connection instantiation to use the `.new_client` pattern should be opened as a separate issue.

Doing the refactoring in another PR also makes sense to me.

By the way, @aidanharan, @wpolicarpo do you know why [this check](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L365) is not present in [execute_procedure](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L170) or [raw_connection_run](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L170)?

mgrunberg: I've reverted the usage of `@connection.active?` but still raising `StatementInvalid` for "failed to execute statement" error.

aidanharan: 1. When you call `connection.active?` the `SELECT 1` query is sent to the database where it is evaluated. The query incurs the cost of the roundtrip and taking up DBMS resourses.

2. The "failed to execute statement" error is thrown (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/0e0da503c0268122f6ff96631ce1ae4e3622f72d/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L365) if the connection returns `false` (https://github.com/rails-sqlserver/tiny_tds/blob/77341a31dbed0299ae6160ec75f00089f95880c7/ext/tiny_tds/client.c#L256). I think that in this situation we can assume the connection is dead. If the connection is dead then I think `ConnectionNotEstablished` is a more appropiate error to throw than `StatementInvalid`, whose naming suggests there is something wrong with the statement rather than the connection.

3. It looks like the [check](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L365) should be present in execute_procedure and raw_connection_run too.

aidanharan: Yes, I was talking about `SQLServerAdapter#active?`. Anyway, don't think we need to call `TinyTds::Client.active?` either as TinyTDS calls it internally when you call `SQLServerAdapter#execute`.

PR #907, opened 2021-04-21
Rails 6.1: Skip test if it causes query plan generation error

aidanharan:
The test `EagerLoadingTooManyIdsTest#test_eager_loading_too_many_ids` is failing on the CI with error:
```
ActiveRecord::StatementInvalid: TinyTds::Error: The query processor ran out of internal resources and could not produce a query plan. This is a rare event and only expected for extremely complex queries or queries that reference a very large number of tables or partitions. Please simplify the query. If you believe you have received this message in error, contact Customer Support Services for more information.
```

This happens if the SQL Server database is unable to handle the large number of values in the IN clause when generating the query plan. This is a known issue with SQL Server (see https://www.mssqltips.com/sqlservertip/5279/sql-server-error-query-processor-ran-out-of-internal-resources-and-could-not-produce-a-query-plan/).

I have found that the test passes if the database compatibility level is 100/110/120 but fails if it's 130/140. I was going to skip the test based on the compatibility level however "it is possible to generate this error during execution due to limitation of server resources like memory, etc" (https://www.mssqltips.com/sqlservertip/5279/sql-server-error-query-processor-ran-out-of-internal-resources-and-could-not-produce-a-query-plan/). 

The issue, when it happens, is with the database instance and not the SQL Server adapter. So by skipping the test
only when the query plan generation fails we can do our best to catch any genuine errors while ignoring a known
SQL Server issue.

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1732,3 +1732,23 @@ class BasePreventWritesLegacyTest < ActiveRecord::TestCase
     end
   end
 end
+
+require "models/citation"
+class EagerLoadingTooManyIdsTest < ActiveRecord::TestCase
+  # This test can fail with error "The query processor ran out of internal resources and could not produce a query plan"
+  # if the SQL Server database is unable to handle the large number of values in the IN clause when generating the query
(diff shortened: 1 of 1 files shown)


aidanharan:
Closing in favor of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/909

PR #930, opened 2021-07-14
Remove QueryMethods patch that is no longer required

aidanharan:
Remove the QueryMethods patch that is no longer required since the `main` branch is now for Rails 6.1+ development.

See https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/activerecord/lib/active_record/relation/query_methods.rb#L1322 for the Rails method that matches the patch being removed.


aidanharan:
Tests will be fixed when https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/931 is merged to `master`.

PR #933, opened 2021-07-19
Apply monkey patches for SQL Server connections only

### LINKED ISSUE #929: Do the monkey patches in this gem interfere with the Rails built-in PostgreSQL adapter?
## Issue

This is more of a question than a reproducible issue.

My team is considering using the [multiple database support](https://guides.rubyonrails.org/active_record_multiple_databases.html) built into Rails to connect our Rails app to two databases simultaneously: a PostreSQL database and a SQL Server database. Does this gem supports such a scenario?

For example, Rails 6.1 supports a `database.yml` like this:

```yaml
production:
  primary:
    adapter: postgresql;
    # postgres config goes here
  secondary:
    adapter: sqlserver
    # sql server config goes here
```

Then one can use [establish_connection](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-establish_connection) to connect to one or the other database on demand. Alternatively, one can declare models using [connects_to](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connects_to) to "bind" models to a certain database.

However, **I see that this gem applies monkey patches to several ActiveRecord internal classes.** Which has me worried. See here:

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/tree/v6.1.0.0/lib/active_record/connection_adapters/sqlserver/core_ext

Some of these patches seem to imply that compatibility with other databases is being removed (notably PostgreSQL in [this example](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/v6.1.0.0/lib/active_record/connection_adapters/sqlserver/core_ext/calculations.rb#L11)) in favor of supporting SQL Server. This is not explicitly stated in the code or documentation, so I can't be sure of that.

**Is it problematic to use the SQL Server adapter and the PostgreSQL adapter in the same Rails application?** It seems like conflicts are a real possibility, because the monkey patches are scoped to affect ActiveRecord as a whole.

## Expected behavior

I'd like like to be able to use the SQL Server adapter and PostgreSQL adapter in the same Rails app without them conflicting with each other. Is it possible that the monkey patches in this gem could somehow be scoped so they only apply based on the database connection?

In other words, in pseudo code:

> if connection.adapter is sqlserver then patched_version else original_version

## Details

- **Rails version**: `6.1.4`
- **SQL Server adapter version**: `6.1.0.0`
- **FreeTDS details**:

 ```
 $ tsql -C
                            Version: freetds v1.2.21
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
  ```

mattbrictson:
Recent versions of Rails support multiple database connections within the same app. It is possible for these connections to use different adapters. For example, one adapter may use SQL Server, and another uses PostgreSQL.

This gem applies some monkey patches to ActiveRecord for SQL Server compatibility. These patches could break other adapters, though, in a multiple-database scenario.

This PR modifies the patches so that they are applied only if the connection is SQL Server. If not, the original ActiveRecord implementation (`super`) is used instead.

Fixes #929

@aidanharan @wpolicarpo what do you think of this approach? Any suggestions on how to test?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## Unreleased
+
+#### Fixed
+
+- [#933](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/933) Conditionally apply SQL Server monkey patches to ActiveRecord so that it is safe to use this gem alongside other database adapters (e.g. PostgreSQL) in a multi-database Rails app
+
+#### Changed
+
+- ...
+
+#### Added
+
+- ...
+
 ## v6.1.0.0
 
 - No changes
--- a/lib/active_record/connection_adapters/sqlserver/core_ext/attribute_methods.rb
+++ b/lib/active_record/connection_adapters/sqlserver/core_ext/attribute_methods.rb
@@ -10,6 +10,8 @@ module ActiveRecord
           private
 
           def attributes_for_update(attribute_names)
+            return super unless self.class.connection.adapter_name == "SQLServer"
+
             super.reject do |name|
               column = self.class.columns_hash[name]
               column && column.respond_to?(:is_identity?) && column.is_identity?
--- a/lib/active_record/connection_adapters/sqlserver/core_ext/calculations.rb
+++ b/lib/active_record/connection_adapters/sqlserver/core_ext/calculations.rb
@@ -10,6 +10,8 @@ module ActiveRecord
         module Calculations
           # Same as original except we don't perform PostgreSQL hack that removes ordering.
           def calculate(operation, column_name)
+            return super unless klass.connection.adapter_name == "SQLServer"
+
             if has_include?(column_name)
               relation = apply_join_dependency
 
@@ -29,6 +31,8 @@ module ActiveRecord
           private
 
           def build_count_subquery(relation, column_name, distinct)
+            return super unless klass.connection.adapter_name == "SQLServer"
+
             super(relation.unscope(:order), column_name, distinct)
           end
         end
--- a/lib/active_record/connection_adapters/sqlserver/core_ext/explain.rb
+++ b/lib/active_record/connection_adapters/sqlserver/core_ext/explain.rb
@@ -9,6 +9,8 @@ module ActiveRecord
           SQLSERVER_STATEMENT_REGEXP = /N'(.+)', N'(.+)', (.+)/
 
           def exec_explain(queries)
+            return super unless connection.adapter_name == "SQLServer"
+
             unprepared_queries = queries.map do |(sql, binds)|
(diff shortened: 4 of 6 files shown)


mattbrictson:
Hmm. Appveyor CI is failing with [one failing test](https://ci.appveyor.com/project/rails-sqlserver/activerecord-sqlserver-adapter/builds/40053460/job/n5phge5t2bii31r3#L369), but that same test fails for me locally on `main`...

On `main`, running locally:

```
Failure:
LogSubscriberTest#test_where_in_binds_logging_include_attribute_names [/Users/mattb/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/bundler/gems/rails-f94f41dd59b5/activerecord/test/cases/log_subscriber_test.rb:252]:
Expected /\["id",\ 1\],\ \["id",\ 2\],\ \["id",\ 3\],\ \["id",\ 4\],\ \["id",\ 5\]/ to match "Developer Load (2.3ms)  EXEC sp_executesql N'SELECT [developers].[id], [developers].[name], [developers].[salary], [developers].[firm_id], [developers].[mentor_id], [developers].[legacy_created_at], [developers].[legacy_updated_at], [developers].[legacy_created_on], [developers].[legacy_updated_on] FROM [developers] WHERE [developers].[id] IN (@0, @1, @2, @3, @4)', N'@0 bigint, @1 bigint, @2 bigint, @3 bigint, @4 bigint', @0 = 1, @1 = 2, @2 = 3, @3 = 4, @4 = 5  [[\"id\", nil], [\"id\", nil], [\"id\", nil], [\"id\", nil], [\"id\", nil]]".
```

aidanharan:
Yeah this fix looks fine to me. The `LogSubscriberTest#test_where_in_binds_logging_include_attribute_names` test should be fixed already by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/931

wpolicarpo:
Also, could you rebase your branch and add a changelog entry?

mattbrictson:
@wpolicarpo thanks for the feedback! I think I've addressed all of your concerns:

- Rebased against main
- Changed `is_a?(SQLServerAdapter)` to `adapter_name == "SQLServer"`
- Added an **Unreleased** section to the CHANGELOG with an entry for this PR

I also squashed this PR down to one commit.

Let me know if there is anything else you need.

wpolicarpo:
All good. Thanks for your contribution.

PR #941, opened 2021-09-01
Fix application name sent to TinyTDS

aidanharan:
Fixes the application name so that the Rails application name can be used for the database connection created by TinyTDS. 

During the refactoring in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/917 the method `configure_application_name` should have been moved from an instance to a class method. This resulted in the application name never getting set to the Rails application name as the call to `configure_application_name` would throw a `NameError` exception, which resulted in the application name being set to `nil`. The TinyTds client would ignore the `nil` and use 'TinyTds' as the application name instead (https://github.com/rails-sqlserver/tiny_tds/blob/e3e79a030271891b11b6e7676382fe31f389b099/lib/tiny_tds/client.rb#L50).

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver_adapter.rb
+++ b/lib/active_record/connection_adapters/sqlserver_adapter.rb
@@ -105,7 +105,23 @@ module ActiveRecord
         end
 
         def config_appname(config)
-          config[:appname] || configure_application_name || Rails.application.class.name.split("::").first rescue nil
+          if self.instance_methods.include?(:configure_application_name)
+            ActiveSupport::Deprecation.warn <<~MSG.squish
+            Configuring the application name used by TinyTDS by overriding the
+            `ActiveRecord::ConnectionAdapters::SQLServerAdapter#configure_application_name`
+            instance method is no longer supported. The application name should configured
+            using the `appname` setting in the `database.yml` file instead. Consult the
+            README for further information."
+            MSG
+          end
+
+          config[:appname] || rails_application_name
+        end
+
+        def rails_application_name
+          return nil if Rails.application.nil?
+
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -119,6 +119,8 @@ def config_timeout(config)
         def config_encoding(config)
           config[:encoding].present? ? config[:encoding] : nil
         end
+
+        def configure_application_name; end
```
wpolicarpo, line 123: Changing this means the README is wrong. I think we should remove this method altogether and move the configuration to `database.yml`.

There are 2 problems:

1. Before removing, we should deprecate its usage so users have time to fix their applications
2. The current behaviour is incorrect and it is failing silently

What I think should be done:

1. Fix `config_appname` so it warns users that their config is wrong if they overwrote `configure_application_name` and suggest them to use `appname` in their `database.yml`
2. Change docs to suggest using `database.yml` instead

lib/active_record/connection_adapters/sqlserver_adapter.rb
```diff
@@ -105,7 +105,23 @@ def dblib_connect(config)
...
+            instance method is no longer supported. The application name should configured
+            using the `appname` setting in the `database.yml` file instead. Consult the
+            README for further information."
+            MSG
+          end
+
+          config[:appname] || rails_application_name
+        end
+
+        def rails_application_name
```
aidanharan, line 121: This was added as if `Rails.application` is `nil` then the result was "NilClass" when `nil` would have been expected.
```
nil.class.name.split("::").first
> "NilClass"
```

PR #957, opened 2021-11-28
use tsql-ttds for structure_load (fixes #956)

### LINKED ISSUE #956: support views in `structure_load`
## Issue

Given an example schema:
```sql
create table posts
(
    ID   int identity,
    Body varchar(max) not null,
)

create view popular_posts
as
select *
from posts
```

`structure_dump` [strips the `Go` statements added in between the various objects](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/main/lib/active_record/tasks/sqlserver_database_tasks.rb#L75). As a result, when loading the objects back into the database via `structure_load`, the following error is thrown.

```sh
$ rails db:setup
...
rails aborted!
ActiveRecord::StatementInvalid: TinyTds::Error: 'CREATE VIEW' must be the first statement in a query batch.
/Users/jonahgeorge/src/funcard/funcard/bin/rails:5:in `<top (required)>'
/Users/jonahgeorge/src/funcard/funcard/bin/spring:10:in `block in <top (required)>'
/Users/jonahgeorge/src/funcard/funcard/bin/spring:7:in `<top (required)>'
```

## Expected behavior

`structure_load` should be capable of loading a `structure.sql` file generated by `structure_dump`.

## Actual behavior

`TinyTds::Error: 'CREATE VIEW' must be the first statement in a query batch.` exception is raised.

## Details

- **Rails version**: `6.1.4`
- **SQL Server adapter version**: `v6.1.2.1`
- **FreeTDS details**:

 ```
$ tsql -C    
                            Version: freetds v1.3.3
             freetds.conf directory: /opt/homebrew/etc
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
  ```

jonahgeorge:
This PR modifies `structure_load` to utilize `tsql-ttds` to load the `structure.sql` file generated by `structure_dump`. With this change, we can remove the gsub to remove `GO` commands which should allow views to be properly imported.

Fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/956

### DIFF
--- a/lib/active_record/tasks/sqlserver_database_tasks.rb
+++ b/lib/active_record/tasks/sqlserver_database_tasks.rb
@@ -72,7 +72,6 @@ module ActiveRecord
 
         dump = File.read(filename)
         dump.gsub!(/^USE .*$\nGO\n/, "")                      # Strip db USE statements
-        dump.gsub!(/^GO\n/, "")                               # Strip db GO statements
         dump.gsub!(/nvarchar\(8000\)/, "nvarchar(4000)")      # Fix nvarchar(8000) column defs
         dump.gsub!(/nvarchar\(-1\)/, "nvarchar(max)")         # Fix nvarchar(-1) column defs
         dump.gsub!(/text\(\d+\)/, "text")                     # Fix text(16) column defs
@@ -80,7 +79,18 @@ module ActiveRecord
       end
 
       def structure_load(filename, extra_flags)
-        connection.execute File.read(filename)
+        server_arg = "-S #{Shellwords.escape(configuration_hash[:host])}"
+        server_arg += ":#{Shellwords.escape(configuration_hash[:port])}" if configuration_hash[:port]
+        command = [
+          "tsql-ttds",
+          server_arg,
+          "-D #{Shellwords.escape(configuration_hash[:database])}",
+          "-U #{Shellwords.escape(configuration_hash[:username])}",
+          "-P #{Shellwords.escape(configuration_hash[:password])}",
+        ]
+
+        stdout_str, stderr_str, status = Open3.capture3(command.join(" "), stdin_data: File.read(filename))
+        raise "Error loading database: #{stderr_str}" unless status.exitstatus == 0
       end
 
       private


ttilberg:
@jonahgeorge Did you close this for any particular reason? Did you find an alternate solution?

I'm trying to support a join table using a view in my Rails app, and ye ole `'CREATE VIEW' must be the first statement in a query batch.` is in my way after fixing the unquoted `[key]` column name in AR Metadata...

- Did you solve this a different way?
- Did you find other strange issues you had to overcome using SQL Server + structure.sql?
- Did you give up on sqlserver? 🤣

travismiller:
I ran into the same `[key]` and `first statement in a query batch` issues using `db/structure.sql`. These changes worked for me and this _feels_ like the right change since it uses the same toolset for both dump and load.

jonahgeorge:
> @jonahgeorge Did you close this for any particular reason? Did you find an alternate solution?
> 
> I'm trying to support a join table using a view in my Rails app, and ye ole `'CREATE VIEW' must be the first statement in a query batch.` is in my way after fixing the unquoted `[key]` column name in AR Metadata...
> 
> * Did you solve this a different way?
> * Did you find other strange issues you had to overcome using SQL Server + structure.sql?
> * Did you give up on sqlserver? 🤣

It appears I missed this notification way back when 🙈 Sorry about that! I'm operating on a slightly hazy old memory here, so bear with me.

> * Did you solve this a different way?
> * Did you find other strange issues you had to overcome using SQL Server + structure.sql?

I recall having issues with the format generated by `structure_dump` here primarily due to the usage of `GO` statements. I ended up using a `structure.sql` file generated by DataGrip & then partially hand edited as I could not get any CLI tooling to work for my use-case.

> * Did you give up on sqlserver? 🤣

Hah, I did not; however, the current solution is subpar-- We don't redump the structure on migrations and instead rely on only the migration chain. 

---

I may have some time in the coming months to take another look at this; especially, due to motivation to move off of my fork for Rails 7.1 support. I'll post any updates on this PR and potentially will re-open it.

travismiller:
Probably a better place for this, but I'll leave it here for now.

I've also encountered an issue in the dumped structure.sql file where the default NEWID is not being included for UUID primary keys.

jonahgeorge:
A few recent findings:

My initial attempts at using the schema created by `defncopy` were unsuccessful. It appears this tool has a handful of issues including limited support for stored procedures, [improperly escaped names](https://github.com/FreeTDS/freetds/issues/429), etc.

A while back I got around this by using a combination [SQL Scripts > Generate DDL to Clipboard](https://www.jetbrains.com/datagrip/features/generation.html) feature in Jetbrains DataGrip + [`config.active_record.dump_schema_after_migration = false`](https://edgeguides.rubyonrails.org/configuring.html#config-active-record-dump-schema-after-migration) but really disliked not having an open source or scriptable way to do this.

---

In looking for more recent tools for schema dumping I came across [`microsoft/mssql-scripter`](https://github.com/microsoft/mssql-scripter)-- this is a thin Python wrapper around the [`microsoft/sqltoolsservice`](https://github.com/microsoft/sqltoolsservice) C# JSONRPC server. 

A quick test of commands below (with the patches in this PR & some minimal manual editing of the generated) applied successfully; however, as the migrations table is not populated is causes issues with running tests. I'll have to look at how the Postgres adapter includes data for a subset of tables.

```sh
pip install mssql-scripter

mssql-scripter \
  -S localhost \
  -U "user" \
  -P "password" \
  -d "database" \
  --display-progress \
  --enable-toolsservice-logging > db/structure.sql

rails db:create && rails db:schema:load
```

Potentially a path forward for the library would be to depend on one of the projects listed above. Alternatively, I was able to track down the majority of the schema generation logic to the following package but it would be a mountain of work to re-implement in Ruby.

https://github.com/microsoft/sqlmanagementobjects/blob/main/src/Microsoft/SqlServer/Management/SqlScriptPublish

Michoels:
The FreeTDS guys have a fix for `defncopy` on the way: https://github.com/FreeTDS/freetds/issues/429#issuecomment-2028871684

Issue #960, opened 2021-12-16
Rails7 is out


net1957:
Do you have any plan to support it ?

PR #961, opened 2021-12-22
Start Rails 7.0 support

mgrunberg:
This PR does almost the minimum changes to be able to run tests with Rails 7. It's the starting point to work on Rails 7 support.

The PR:
- Drops support of ruby 2.5 and 2.6. Rails 7 requires ruby >= 2.7.0
- Adds benchmark-ips to Gemfile. Is required by ActiveRecord attribute encryption testcases
- Inline `quote` and `type_cast` methods. Rails did that so `_quote` method is not called anymore. Tests didn't run because the query to fetch column definitions was failing because of bad quoting

```
Whitout inlineing quote method
EXEC sp_executesql 'SELECT [posts].* FROM [posts] WHERE [posts].[author_id] IN (@0, @1)', '@0 bigint, @1 bigint, @2 bigint', @0 = 1, @1 = 2,  [["author_id", nil], ["author_id", nil], ["author_id", nil]]

After
EXEC sp_executesql N'SELECT [posts].* FROM [posts] WHERE [posts].[author_id] IN (@0, @1)', N'@0 bigint, @1 bigint', @0 = 1, @1 = 2  [["author_id", nil], ["author_id", nil]]
```

The PR also removes two deprecation warnings. It's not required to start working on Rails 7 support, could be a different PR, but reduces the output a lot.

Related with #960


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.5.9, 2.6.7, 2.7.3, 3.0.1]
+        ruby: [2.7.3, 3.0.1]
```
aidanharan, line 16: Should this not be the latest Ruby 3.0 version, which is 3.0.3?

aidanharan, line 16:
```suggestion
        ruby: [2.7.5, 3.0.3]
```

mgrunberg:
@aidanharan I had doubts about upgrading 2.7.3 to 2.7.5 and 3.0.1 to 3.0.3. But shouldn't do any harm.

.github/workflows/ci.yml
aidanharan: @mgrunberg My mistake. It looks like the Docker images have to updated first (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4609524313?check_suite_focus=true). Could you revert this change?

PR #968, opened 2021-12-27
[Rails 7] Define adapter type maps statically

mgrunberg:
Related to https://github.com/rails/rails/pull/42773
Follows the same work done in [this commit](https://github.com/rails/rails/commit/d79fb963603658117fd1d639976c375ea2a8ada3). 

Also fixes by updating the testcase.
```
QueryCacheTest#test_query_cached_even_when_types_are_reset_coerced:
2021-12-24T09:02:33.8588183Z NoMethodError: undefined method `reload_type_map' for #<ActiveRecord::ConnectionAdapters::SQLServerAdapter version: 7.0.0.0-beta.1, mode: dblib, azure: false>
2021-12-24T09:02:33.8590358Z     activerecord-sqlserver-adapter/test/cases/coerced_tests.rb:1133:in `block in test_query_cached_even_when_types_are_reset_coerced'
2021-12-24T09:02:33.8593111Z     rails (5850a6592ff1) activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb:61:in `cache'
2021-12-24T09:02:33.8595184Z     rails (5850a6592ff1) activerecord/lib/active_record/query_cache.rb:11:in `cache'
2021-12-24T09:02:33.8596652Z     activerecord-sqlserver-adapter/test/cases/coerced_tests.rb:1128:in `test_query_cached_even_when_types_are_reset_coerced'
----
LogSubscriberTest#test_binary_data_is_not_logged:
ActiveRecord::StatementInvalid: TinyTds::Error: Implicit conversion from data type nvarchar(max) to varbinary(max) is not allowed. Use the CONVERT function to run this query.
----
Failure:
TimePrecisionTest#test_time_precision_is_truncated_on_assignment_coerced [/activerecord-sqlserver-adapter/test/cases/coerced_tests.rb:1433]:
Expected: 123457000
  Actual: 123456000
```

Before (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4792357743?check_suite_focus=true):
```
7744 runs, 21077 assertions, 75 failures, 45 errors, 43 skips
```

After (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4793438242?check_suite_focus=true):
```
7744 runs, 25975 assertions, 4 failures, 2 errors, 43 skips
```


test/cases/coerced_tests.rb
```diff
@@ -1130,7 +1130,7 @@ def test_query_cached_even_when_types_are_reset_coerced
       Task.find(1)
 
       # Preload the type cache again (so we don't have those queries issued during our assertions)
-      Task.connection.send(:reload_type_map)
+      Task.connection.send(:reload_type_map) if Task.connection.respond_to?(:reload_type_map, true)
```
aidanharan, line 1133: Can `Task.connection.send(:reload_type_map)` not simply be removed since `respond_to?` will always be `false` for the SQL Server adapter in Rails 7?

mgrunberg: It's true that the whole line can be removed. I choose to add the `respond_to?` check to honor the description of the test `Same as original test except that we expect one query to be performed...`. This way the body of the test is Rails 7 + the extra assertion. Easier to maintain in my opinion. What do you think?

mgrunberg: Removed the call and changed the test comment

PR #973, opened 2021-12-30
[Rails 7] Coerce tests that use upsert

mgrunberg:
SQL Server does not support upsert yet. There are 7 tests that do not check if the driver support it. This PR coerce these tests.

Before (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4674744906?check_suite_focus=true):
```
7739 runs, 21036 assertions, 94 failures, 60 errors, 43 skips
```

After (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4675657652?check_suite_focus=true)
```
7739 runs, 21044 assertions, 94 failures, 53 errors, 43 skips
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -2003,3 +2003,56 @@ class QueryLogsTest < ActiveRecord::TestCase
     end
   end
 end
+
+# SQL Server does not support upsert yet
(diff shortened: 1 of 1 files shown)


aidanharan:
@mgrunberg PR looks good but could you squash the commits?

PR #982, opened 2022-01-04
[Rails 7] Coerce encrypted with limit test

mgrunberg:
Fixes:
````
Failure:
ActiveRecord::Encryption::EncryptableRecordTest#test_0032_validate column sizes [/usr/local/bundle/bundler/gems/rails-5850a6592ff1/activerecord/test/cases/encryption/encryptable_record_test.rb:254]:
Expected true to be nil or false

rails test usr/local/bundle/bundler/gems/rails-5850a6592ff1/activerecord/test/cases/encryption/encryptable_record_test.rb:252
````

SQL Server string column limit is 4000. Other adapters use 255. Reimplement test to produce an invalid string considering SQL Server limit

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -2155,3 +2155,15 @@ class InsertAllTest < ActiveRecord::TestCase
     assert_equal %w[ REWORK ], result.pluck("name")
   end
 end
+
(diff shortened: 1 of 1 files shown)


aidanharan:
@mgrunberg Could you merge `main` into the branch and squash the commits?

mgrunberg:
@aidanharan done

PR #987, opened 2022-01-06
[Rails 7] Reimplement test that build queries with group by and then call `all`

mgrunberg:
Rails introduced the ability to call delete_all and update_all on scopes
that use having and group by doing a subquery. See [here](https://github.com/rails/rails/commit/989d53ba3b3b43d3b1a179a146be9244e1c2537e)
and [here](https://github.com/rails/rails/commit/4acb6660e265fd40b38e301fbd15909574d6a59e)

To test this they build scopes like
`Post.group('posts.id').joins(:comments).having('count('comments.id') >
2').all`. SQL Server adapter doesn't support these. Results in
```
ActiveRecord::StatementInvalid: TinyTds::Error: Column 'posts.author_id'
is invalid in the select list because it is not contained in either an
aggregate function or the GROUP BY clause.
```

Coerce and reimplement the test to use subqueries.

Before (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4674744906?check_suite_focus=true):
```
7739 runs, 21036 assertions, 94 failures, 60 errors, 43 skips
```

After (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/runs/4732617519?check_suite_focus=true):
```
7743 runs, 21058 assertions, 94 failures, 58 errors, 43 skips
```

### DIFF
--- a/test/cases/coerced_tests.rb
+++ b/test/cases/coerced_tests.rb
@@ -1090,6 +1090,24 @@ class UpdateAllTest < ActiveRecord::TestCase
     _(david.reload.name).must_equal "David"
     _(mary.reload.name).must_equal "Test"
   end
+
+  # SQL Server doesn't support queries like SELECT table.* FROM table GROUP BY table.id.
+  # These kind of query is generated to get the affeted records.
+  # Same as original but moving conflicting queries into subqueries
+  coerce_tests! :test_update_all_with_group_by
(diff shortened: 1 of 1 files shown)


aidanharan:
Fixed the tests in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/991 by selecting the required columns rather than using subqueries.

PR #1002, opened 2022-02-08
[Rails 7.0] Fix support for different index types

aidanharan:
Fix support for different index types. 

Index types were working in Rails 6.0 but were broken in Rails 6.1 by https://github.com/rails/rails/pull/39203. The issue was identified in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/999 for Rails 6.1.

This PR is for Rails 7.0. The fix will be backported to the `6-1-stable` branch after it is merged into `main`.

Issue #999, opened 2022-02-07
ActiveRecord::StatementInvalid (TinyTds::Error: Incorrect syntax near the keyword 'clustered'.)


jjimenez:
## Issue
When trying to create a clustered index, the wrong sql statement is generated.  We do this when we create reporting tables from our rails data.  We drop and recreate tables and re-create indexes that were removed.

## Expected behavior
connection.add_index "testings", "last_name", type: :clustered
should generate sql: "CREATE CLUSTERED INDEX [index_testings_on_last_name] ON [testings] ([last_name])"

## Actual behavior
sql generated is "CREATE INDEX clustered [index_testings_on_last_name] ON [testings] ([last_name])"

## How to reproduce

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "tiny_tds"
  gem "activerecord", "6.1.0"
  gem "activerecord-sqlserver-adapter", "6.1.0"
end

require "active_record"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(
  adapter:  "sqlserver",
  timeout:  5000,
  pool:     100,
  encoding: "utf8",
  database: "activerecord_unittest",
  username: "rails",
  password: "",
  host:     "localhost",
  port:     1433,
)
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  drop_table :testings rescue nil
  
  create_table :testings, force: true  do |t|
    t.column :foo, :string, limit: 100
  end
end

class Testings < ActiveRecord::Base
end

class TestBugTest < Minitest::Test
  def test_index_type
    Testings.connection.add_index "testings", "id", type: :clustered
  end
end

## Details

- **Rails version**: `6.1.0`
- **SQL Server adapter version**: `6.1.0`
- **TinyTDS version**: `2.1.5`
- **FreeTDS details**:

  ```
  run `tsql -C` and paste here the output.

Compile-time settings (established with the "configure" script)
                            Version: freetds v1.3.7
             freetds.conf directory: /opt/homebrew/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: yes
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes

  ```

[...]

PR #1006, opened 2022-02-10
[Rails 6.1] Fix support for different index types

aidanharan:
Back-port of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1002 into the `6-1-stable` branch.

Fixes the https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/999 issue that was initially reported for Rails 6.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+#### Fixed
+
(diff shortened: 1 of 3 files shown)


aidanharan:
Windows CI failing because of issue with the `pg` gem. See https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1007

PR #1019, opened 2022-07-20
fix-change-column-default

### LINKED ISSUE #1018: change_column_default cannot drop default constraint
## Issue
change_column_default and similar methods not working as intended

## Expected behavior

according to the documentation
[change_column_default](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/change_column_default)(:users, :email, nil)

`change_column_default("SomeTable", "SomeColumn", nil)`

should drop the default constraint on SomeColumn.

## Actual behavior

`ALTER TABLE [SomeTable] ADD CONSTRAINT DF_SomeTable_SomeColumn DEFAULT NULL FOR [SomeColumn]`

gets executed which results in a `(Null)` default constraint.

## How to reproduce

use
`change_column_default("SomeTable", "SomeColumn", nil)`

## Details

- **Rails version**: `7.0.1`
- **SQL Server adapter version**: `7.0.0.0`

nemesit:
Would fix #1018

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -179,7 +179,7 @@ module ActiveRecord
 
           remove_default_constraint(table_name, column_name)
           default = extract_new_default_value(default_or_changes)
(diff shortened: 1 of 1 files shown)


aidanharan:
@nemesit Would you be able to add tests to the PR? The default constraint should be dropped when `nil` is passed rather than not altered. Could you confirm in your tests that this happens?

PR #1024, opened 2022-09-29
Add dbconsole support to adapter

gmcgibbon:
Related to https://github.com/rails/rails/pull/46093

Adds dbconsole method to be used when invoking bin/rails dbconsole.

Moves dbconsole logic related to SQL Server in Rails to the adapter. This connection code and corresponding test was in the dbconsole command in railties.

cc @paarthmadan


wpolicarpo:
Thanks for that.

Issue #1032, opened 2023-02-13
Procs in schema default doesn't work since Ruby 3.2.0


ryntluka:
## Issue
SQLServer Adapter from Ruby 3.2.0 doesn't work correctly with procs in shema as defaults.

## Expected behavior
For example when using custom defaults with calls to DB like `NEWID()` for uuid `rails db:schema:dump` generates column with default proc and it would normally work correctly when running `rails db:schema:load`.

## Actual behavior
Now though, when dumping the schema and running load on it it results in an error:
```
NoMethodError: undefined method `=~' for #<Proc:..
```

## How to reproduce
To be more concrete adding a column in migrations like this:
```ruby
add_column :users, :uuid, :uuid, default: 'NEWID()'
```

writes out into `schema.rb` file in the following format:

```ruby
t.uuid "uuid", default: -> { "newid()" }, null: false
```

and when parsing this schema, rails fails with error on the proc.

## Details

I actually managed to pinpoint the source of the problem to be in calling method `ActiveRecord::ConnectionAdapters::SQLServer::Quoting#quote_default_expression`. There is a condition that calls `=~` on proc like this:

```ruby
if cast_type.type == :uuid && value =~ /\(\)/
```

In previous versions of ruby this would work correctly (although it would always return nil for proc) but since 3.2.0 this [was removed](https://bugs.ruby-lang.org/issues/15231) and it causes the `NoMethodError`.

PR #1033, opened 2023-02-27
Support proc default values in ruby 3.2 

mgrunberg:
Fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1032
```
NoMethodError: undefined method `=~' for #<Proc:..
```

As [the issue](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1032) explains `=~` [was removed](https://bugs.ruby-lang.org/issues/15231) from Proc.

`Proc` are handled by [this line of code](https://github.com/rails/rails/blob/v7.0.4.2/activerecord/lib/active_record/connection_adapters/abstract/quoting.rb#L96)

The solution is to add a check and call `~=` only when `value` is a `String`.

I took the oportunity to follow rubocop performance suggestions
```
Performance/StringInclude: Use String#include? instead of a regex match
with literal-only pattern.
```
Got the idea from postgresql adapter
(https://github.com/rails/rails/commit/5726b1d1d7665c33830ad114f45258ecf772b7b6#diff-d8e4ab5c8c59df4fc1bc85457af0d07b88bd842a287233617a97ad2436fcb842R120)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 #### Fixed
 
 - [#1029](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1029) Handle views defined in other databases.
+- [#1033](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1033) Support proc default values in ruby 3.2.
+
 
 #### Changed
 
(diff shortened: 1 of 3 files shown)


aidanharan:
@mgrunberg Could you make the following changes:

1. Remove Ruby 3.2 from the CI matrix as the CI doesn't support it yet. I've opened this as https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1034
2. Update the change-log with your fix.
3. Squish your commits.

Thanks Aidan

Issue #950, opened 2021-10-11
Nil variable @version_year is causing error when Rails try to load schema on the test database


theo-bittencourt:
## Issue

Cannot setup test db properly because `db:schema:load` is raising  `NoMethodError: undefined method `<' for nil:NilClass`.

In the method `ConnectionAdapters::SQLServer::SchemaStatements.drop_table`, the variable `@version_year` is not defined yet at this point.

## Expected behavior

Should load schema on test db.

## Actual behavior

`rails db:schema:load RAILS_ENV=test ` or `rails db:test:prepare` are  raising  `NoMethodError: undefined method `<' for nil:NilClass`.

## How to reproduce

Run `rails db:schema:load RAILS_ENV=test ` or `rails db:test:prepare`.

## Details

- **Rails version**: `6.1.4.1`
- **SQL Server adapter version**: `6.1.2.1`
- **TinyTDS version**: `2.1.5`
- **FreeTDS details**:

```
Compile-time settings (established with the "configure" script)
                            Version: freetds v1.3.2
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: yes
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
```

## Stacktrace

```
NoMethodError: undefined method `<' for nil:NilClass
.../ruby/3.0.2/lib/ruby/gems/3.0.0/gems/activerecord-sqlserver-adapter-6.1.2.1/lib/active_record/connection_adapters/sqlserver/schema_statements.rb:29:in `drop_table'
.../ruby/3.0.2/lib/ruby/gems/3.0.0/gems/activerecord-6.1.4.1/lib/active_record/connection_adapters/abstract/schema_statements.rb:317:in `create_table'
.../ruby/3.0.2/lib/ruby/gems/3.0.0/gems/activerecord-sqlserver-adapter-6.1.2.1/lib/active_record/connection_adapters/sqlserver/schema_statements.rb:12:in `create_table'
.../ruby/3.0.2/lib/ruby/gems/3.0.0/gems/activerecord-6.1.4.1/lib/active_record/migration.rb:929:in `block in method_missing'
[...]

PR #1039, opened 2023-03-08
Fix hook method that allows custom connection configuration

aidanharan:
The method `configure_connection` was meant to be a hook method to allow custom connection configuration. However, the method was implemented in the adapter to perform the default connection configuration. I have extracted the default connection configuration to it's own method so that the hook method is available again for custom configuration.

This allows users to implement `configure_connection` without having to re-implement the default configuration or call super to call the adapter's implementation of the method.

This fixes issue https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/950

Note: The fix will later be backported to the `6-1-stable` branch.


aidanharan:
@owst Would you be able to review this PR and tell me if it will solve #950 for you? Thanks

owst:
The change looks reasonable to me - thanks!

Issue #1050, opened 2023-04-18
7.0.2.0 identity insert issue


rgeerts:
## Issue
I just upgraded from 7.0.1.0 to 7.0.2.0 and I am trying to run seed some data in a database that calls 
exec_insert in activerecord-sqlserver-adapter-7.0.2.0/lib/active_record/connection_adapters/sqlserver/database_statements.rb
with a query that looks something like "insert into [schema_x].[table_y](id, value) values(1,'abc')". unfortunately, I am getting an error "ActiveRecord::ActiveRecordError: IDENTITY_INSERT could not be turned ON for table [table_y]"

I put in a quick monkey patch that looked like this and the things I was working on started working.

```
def query_requires_identity_insert?(sql)
          return false unless insert_sql?(sql)
          raw_table_name = get_raw_table_name(sql)
          id_column = identity_columns(raw_table_name).first
          #id_column && sql =~ /^\s*(INSERT|EXEC sp_executesql N'INSERT)[^(]+\([^)]*\b(#{id_column.name})\b,?[^)]*\)/i ? SQLServer::Utils.extract_identifiers(raw_table_name).object : false
          id_column && sql =~ /^\s*(INSERT|EXEC sp_executesql N'INSERT)[^(]+\([^)]*\b(#{id_column.name})\b,?[^)]*\)/i ? raw_table_name : false
end
```

## Expected behavior
It looks like it is generating sql that looks like this

"SET IDENTITY_INSERT [table_y] ON" and I would expect it to generate code that looks like "SET IDENTITY_INSERT [schema_x].[table_y] ON"

Our database has many schemas 
Our model looks like
class TableY
  self.table_name = "schema_x.table_y"

...
end

## Details

- **Rails version**: `7.0.4.3`
- **SQL Server adapter version**: `7.0.2.0`
- **TinyTDS version**: `2.1.5`
- **FreeTDS details**:

  ```
  Compile-time settings (established with the "configure" script)
                            Version: freetds v1.3.13
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: yes
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.3
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
  ```

PR #1053, opened 2023-04-20
Fix insertion of records to non-default schema table using raw SQL

aidanharan:
Fix the insertion of records to non-default schema table that has identity column using raw SQL. This fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1050

If you tried to insert records into a table that has an identity column (`id` in this case) using `exec_insert`:

`connection.exec_insert("insert into [test].[aliens] (id, name) VALUES(1, 'Trisolarans'), (2, 'Xenomorph')"`

Then the adapter would first need to turn on `IDENTITY_INSERT` for the table first. However, instead of running:

```
SET IDENTITY_INSERT [test].[aliens] ON;
```

The adapter was ignoring the table schema and trying to turn on `IDENTITY_INSERT` for the table in the default schema.

```
SET IDENTITY_INSERT [aliens] ON;
```

If a table of the same name did not exist in the default schema then an error would be thrown by `SET IDENTITY_INSERT`. If table of the same name in the default schema did exist then the `INSERT` query would fail as it would be unable to insert into the identity column.

_Note: This is not a new issue. The issue was present in at least v7.0.2.0/7.0.1.0/6.1.3.0 of the gem._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Fixed
 
 - [#1052](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1052) Ignore casing of VALUES clause when inserting records using SQL
+- [#1053](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1053) Fix insertion of records to non-default schema table using raw SQL
 
 ## v7.0.2.0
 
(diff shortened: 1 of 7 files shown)


aidanharan:
@rgeerts Would you mind reviewing this PR? Thanks Aidan

rgeerts:
I will look at it and give it a try to see if it works in my situation

PR #1054, opened 2023-05-22
Apply monkey patches for SQL Server connections only (#933)

### LINKED ISSUE #929: Do the monkey patches in this gem interfere with the Rails built-in PostgreSQL adapter?

djezzzl:
This is a cherry-picked fix that I'm kindly asking you to release for the 6.0+ version. 

**The original message**

Recent versions of Rails support multiple database connections within the same app. It is possible for these connections to use different adapters. For example, one adapter may use SQL Server, and another uses PostgreSQL.

This gem applies some monkey patches to ActiveRecord for SQL Server compatibility. These patches could break other adapters, though, in a multiple-database scenario.

This commit modifies the patches so that they are applied only if the connection is SQL Server. If not, the original ActiveRecord implementation (`super`) is used instead.

Fixes #929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Unreleased
 
-[#933](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/933) Conditionally apply SQL Server monkey patches to ActiveRecord so that it is safe to use this gem alongside other database adapters (e.g. PostgreSQL) in a multi-database Rails app
+#### Fixed
+
+[#1054](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1054) Conditionally apply SQL Server monkey patches to ActiveRecord so that it is safe to use this gem alongside other database adapters (e.g. PostgreSQL) in a multi-database Rails app
 
 ## v6.0.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+[#933](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/933) Conditionally apply SQL Server monkey patches to ActiveRecord so that it is safe to use this gem alongside other database adapters (e.g. PostgreSQL) in a multi-database Rails app
+
 ## v6.0.2
 
 #### Fixed
--- a/lib/active_record/connection_adapters/sqlserver/core_ext/attribute_methods.rb
+++ b/lib/active_record/connection_adapters/sqlserver/core_ext/attribute_methods.rb
@@ -10,6 +10,8 @@ module ActiveRecord
           private
 
           def attributes_for_update(attribute_names)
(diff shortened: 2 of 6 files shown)


djezzzl:
Hi @aidanharan, @metaskills, @annaswims, @wbond, @wpolicarpo 👋 

Thank you for the fantastic gem!

Could you please release this fix to the 6.0+ version, too?

djezzzl:
Thank you very much, @aidanharan! 

Have a great week!

Issue #1056, opened 2023-05-29
Tests failing following release of Rails v7.0.5


aidanharan:
## Issue
The CI tests are failing following the release of Rails v7.0.5. The issue is caused by https://github.com/rails/rails/pull/46110

## Expected behavior
Tests should pass.

## Actual behavior
When the test database schema is being created the adapter is sending invalid SQL to MSSQL which results in the error message:

```TinyTds::Error: Column, parameter, or variable #4: Cannot specify a column width on data type datetime. (ActiveRecord::StatementInvalid)```

The issue is with the SQL used to generate data-time columns. For the `aircraft` table (https://github.com/rails/rails/blob/v7.0.5/activerecord/test/schema/schema.rb#L57) the difference is as follows:

**Rails v7.0.4**
```sql
CREATE TABLE [aircraft] ([id] bigint NOT NULL IDENTITY(1,1) PRIMARY KEY, [name] nvarchar(4000), [wheels_count] integer DEFAULT 0 NOT NULL, [wheels_owned_at] datetime2(6), [manufactured_at] datetime DEFAULT CURRENT_TIMESTAMP)
```
 
**Rails v7.0.5**
```sql
CREATE TABLE [aircraft] ([id] bigint NOT NULL IDENTITY(1,1) PRIMARY KEY, [name] nvarchar(4000), [wheels_count] integer DEFAULT 0 NOT NULL, [wheels_owned_at] datetime(6), [manufactured_at] datetime(6) DEFAULT CURRENT_TIMESTAMP)
```

Columns of type `datetime` cannot have precision. 

The SQL generated for Rails v7.0.5 for the `aircraft` table has 2 issues:
1. The `wheels_owned_at` column type has changed from `datetime2(6)` to `datetime(6)`.
2. The `manufactured_at` column type has changed from `datetime` to `datetime(6)`.

## How to reproduce
Run the CI tests.

## Details

- **Rails version**: `7.0.5`
- **SQL Server adapter version**: `7.0.1.0`

PR #990, opened 2022-01-10
[Rails 7] Fix test now rails set precision 6 by default to datetime

mgrunberg:
Requires https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/968

Fix test that got broken because [rails set precision 6 by default to datetime](https://github.com/rails/rails/commit/c2a6f618d22cca4d9b7be7fa7652e7aac509350c)

Fixes:
```
Failure:
SchemaDumperTestSQLServer#test_0002_sst_datatypes_migration [/home/matias/dev/workspace/activerecord-sqlserver-adapter-dev-box/activerecord-sqlserver-adapter/test/cases/schema_dumper_test_sqlserver.rb:70]:
Expected: "datetime"
  Actual: "datetime2(6)"
--------------------
Failure:
ActiveRecord::Migration::ChangeSchemaTest#test_add_column_with_postgresql_datetime_type [/home/matias/.rvm/gems/ruby-2.7.4/bundler/gems/rails-5850a6592ff1/activerecord/test/cases/migration/change_schema_test.rb:308]:
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"datetime(6)"
+"datetime2(6)"
--------------------
Failure:
ActiveRecord::Migration::ChangeSchemaTest#test_change_column_with_timestamp_type [/home/matias/.rvm/gems/ruby-2.7.4/bundler/gems/rails-5850a6592ff1/activerecord/test/cases/migration/change_schema_test.rb:345]:
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"datetime(6)"
+"datetime"
```

PR #1057, opened 2023-05-30
Set precision 6 by default for timestamps

aidanharan:
`Timestamp` columns should use precision 6 by default. This change was included in Rails 7.0.5 by https://github.com/rails/rails/pull/46110

The previous change to default `datetime` columns to use precision 6 was implemented in the adapter by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/990

This fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - [#1052](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1052) Ignore casing of VALUES clause when inserting records using SQL
 - [#1053](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1053) Fix insertion of records to non-default schema table using raw SQL
 
+#### Changed
+
+- [#1057](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1057) Set precision 6 by default for timestamps.
+
 ## v7.0.2.0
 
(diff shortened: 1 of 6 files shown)


aidanharan:
@mgrunberg Could you review this please? If it's okay then I think you can close https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1058

mgrunberg:
@aidanharan Sure, I'll review it later today/tomorrow.

mgrunberg:
@aidanharan In general LGTM.

My only concern, which is not a blocker, is dumping existing `datetime` columns. When you have a database with a `datetime` column, you dump it (`rails db:schema:dump`) and then load schema (`rails db:schema:load`), is the result a `datetime` column?

I couldn't check that. But it seems that [datetime_nil_precision_col](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/f63c90f099e1bdc16f66f830e817c588c35e05aa/test/cases/schema_dumper_test_sqlserver.rb#L85
) and [datetime_col](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/f63c90f099e1bdc16f66f830e817c588c35e05aa/test/cases/schema_dumper_test_sqlserver.rb#L86) are equally dumped. That makes me think that my expectations of dumping & loading are not met.

Sorry if I'm not making myself clear.

mgrunberg:
@aidanharan amazing! I don't have other concerns. LGTM

PR #1058, opened 2023-06-04
Force Rails < 7.0.5

mgrunberg:
I think we should do this until https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1056 is fixed.


mgrunberg:
Closing this PR since #1057 is not a draft anymore

Issue #1042, opened 2023-03-13
Rails 7 fails with Enum columns


KDGundermann:
## Issue
Migrating my old application to Rails 7 my tests fails on accessing the enum columns

## Expected behavior
When accessing an enum column which is defined as:
`enum orderstate: { open: "O", closed: "C" }`
either order.open? or order.closed? should be truthy

## Actual behavior
accessing an enum column always return false 

## How to reproduce
[ActiveRecord Enum Test](https://github.com/KDGundermann/ar_enums)

```
class Datatypes < ActiveRecord::Base
  self.table_name = 'sst_datatypes'

  enum tinyint: { andromeda: 10 , bootes: 20 }
  enum char_10: { alpha: "A", beta: "B" }
end

class EnumTest < Minitest::Test
  def test_enum
    r = Datatypes.new

    r.andromeda!
    r.alpha!

    assert r.andromeda?
    refute r.bootes?

    assert r.alpha?
    refute r.beta?
  end
end
```

## Details

- **Rails version**: 7.0.4.2
-  **Ruby version**: 3.1.3
- **SQL Server adapter version**: 7.0.0.0
- **TinyTDS version**: 2.1.5
- **FreeTDS details**:

  ```
  run `tsql -C` and paste here the output.

  Compile-time settings (established with the "configure" script)
                            Version: freetds v1.3.6
             freetds.conf directory: /usr/local/etc
     MS db-lib source compatibility: no
        Sybase binary compatibility: no
                      Thread safety: yes
                      iconv library: yes
                        TDS version: 7.4
                              iODBC: no
                           unixodbc: no
              SSPI "trusted" logins: no
                           Kerberos: no
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes

  ```

PR #1059, opened 2023-06-04
Fix enums defined on string columns

mgrunberg:
Built on top of #1058 to run specs

Fixes https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1042

varchar, varchar_max and text sql server types extends from char
(`ActiveRecord::ConnectionAdapters::SQLServer::Type:Char`).

`ActiveRecord::ConnectionAdapters::SQLServer::Type::Char.serialize` return an `ActiveRecord::ConnectionAdapters::SQLServer::Type::Data` instance.

`ActiveRecord::ConnectionAdapters::SQLServer::Type::Data` defines
```
def eql?(other)
  self.class == other.class && value == other.value
end
```

Let's imagine the following ActiveRecord definition

```
class MyRecord < ActiveRecord::Base
  enum release: { alpha: "A", beta: "B" }
end
```

When we call `my_record.alpha?` it compares `release` serialized value with the
string "A". It will always be false because `ActiveRecord::ConnectionAdapters::SQLServer::Type::Data != String`.

This commit reimplements `eql?` to deal with string arguments.


aidanharan:
Also, the https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1058 changes should be removed from this PR.

mgrunberg:
@aidanharan done. Please review again

PR #1072, opened 2023-08-04
Improve performance of view default function lookup

albus522:
We are reading from a database outside our control that uses some absolutely massive views (over 500 columns). This makes loading the column definition significantly faster.


aidanharan:
Hi @albus522, what version of Rails are you currently using this patch on? If you are using it on Rails 7.0 then could you make the MR for the `7-0-stable` branch? The `main` branch is being used for Rails 7.1 development.

albus522:
I can switch it to `7-0-stable`. I figured it would start on `main` and be patched, but I am good with whatever works best for you.

aidanharan:
Merged into main in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1074

PR #1073, opened 2023-08-08
Improve performance of view default function lookup

albus522:
`7-0-stable` version of #1072


aidanharan:
Could you add the following to the `CHANGELOG.md`? 

```
## Unreleased

#### Changed

- [#1073](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1073) Improve performance of view default function lookup
```

PR #1111, opened 2023-10-10
Fix Rails application name and SQL Server check 

aovertus:
Hi,

I've been trying to use the `main` version of the adapter on our project and can't make it running yet.
Here are 2 issues I've found so far.

1. Small bug

`rails_application_name` is called on the instance level instead of class

2. Issue with `sqlserver?` declaration

```
3.2.0 :001 > User.where(id: 1).exists?
/Users/alexandreovertus/.rvm/gems/ruby-3.2.0/bundler/gems/activerecord-sqlserver-adapter-d1a3b1dd1977/lib/active_record/connection_adapters/sqlserver/core_ext/finder_methods.rb:14:in `construct_relation_for_exists’: undefined method `sqlserver?' for #<ActiveRecord::ConnectionAdapters::Mysql2Adapter:0x0000000117eaf020 @transaction_manager=#<ActiveRecord::ConnectionAdapters::TransactionManager:0x00000001162e0100 @stack=[], @connection=#
```

`sqlserver?` is unknown from the other adapter. As I understand the goal of defining a method to easily detect which adapter is currently used rather than relying on the name, its not known by the abstracted layer. Ideally the abstract layer would define a method dynamically based on the name of the adapter, in the meantime we can explicitly add a method on the abstract module to make it shared.

Let me know if it makes sense as if its easier to share those notes or make a useable PR to contribute.

Thanks for your support and help with this library 🎉

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/transaction.rb
+++ b/lib/active_record/connection_adapters/sqlserver/transaction.rb
@@ -5,14 +5,12 @@ require "active_record/connection_adapters/abstract/transaction"
 module ActiveRecord
   module ConnectionAdapters
     module SQLServerTransaction
+      delegate :sqlserver?, to: :connection, prefix: true
+
       private
 
-      def sqlserver?
-        connection.respond_to?(:sqlserver?) && connection.sqlserver?
-      end
-
       def current_isolation_level
-        return unless sqlserver?
+        return unless connection_sqlserver?
 
         level = connection.user_options_isolation_level
         # When READ_COMMITTED_SNAPSHOT is set to ON,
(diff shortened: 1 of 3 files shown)


lib/active_record/connection_adapters/sqlserver/transaction.rb
```diff
@@ -7,10 +7,6 @@ module ConnectionAdapters
     module SQLServerTransaction
       private
 
-      def sqlserver?
```
aidanharan, line 10: Think you will need to keep this. Or maybe replace it with:
```
delegate :sqlserver?, to: connection
```

aovertus: Hi @aidanharan I added your suggestion with the `prefix: true` option to help readability as well as tracking where the method is coming from at the transaction level. 

I've tested this locally on our app, it seem to work properly, I did not get the chance to setup the test env from the contributor manual. I'll try to do this this week.

Let me know if there is anything else.

aidanharan:
@aovertus Thanks for the PR! Great to fix the issues you found as soon as possible.

PR #1143, opened 2023-12-22
Fix precision handling for datetimeoffset migration

kiyot:
Resolved an issue where the specified precision for datetimeoffset was not being utilized and was consistently set to 7.
The fix ensures that the custom precision is now correctly applied.


aidanharan:
@kiyot Thanks once again!

PR #1145, opened 2023-12-22
Ensure correct order of COLLATE and NOT NULL in CREATE TABLE statements

kiyot:
COLLATE must appear before NOT NULL for SQL Server

https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql


aidanharan:
Thanks for the fix @kiyot. I just changed your PR slightly by combining the test case into an existing test file.

PR #1152, opened 2024-02-13
Fix Composite Key Inserts with Triggers

jonhinson:
This updates the `OUTPUT INSERTED` statement to handle the case of composite primary keys being properly returned when the table also has a trigger.

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -283,7 +283,7 @@ module ActiveRecord
                     id_sql_type = exclude_output_inserted.is_a?(TrueClass) ? "bigint" : exclude_output_inserted
                     <<~SQL.squish
                       DECLARE @ssaIdInsertTable table (#{quoted_pk.map { |subkey| "#{subkey} #{id_sql_type}"}.join(", ") });
-                      #{sql.dup.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk.join(', INSERTED.')} INTO @ssaIdInsertTable"}
+                      #{sql.dup.insert sql.index(/ (DEFAULT )?VALUES/i), " OUTPUT #{ quoted_pk.map { |subkey| "INSERTED.#{subkey}" }.join(", ") } INTO @ssaIdInsertTable"}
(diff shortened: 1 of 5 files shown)


lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -278,13 +278,13 @@ def sql_for_insert(sql, pk, binds, returning)
...
-                    quoted_pk = SQLServer::Utils.extract_identifiers(pk).quoted
+                    quoted_pk = Array(pk).map { |subkey| SQLServer::Utils.extract_identifiers(subkey).quoted }
 
                     id_sql_type = exclude_output_inserted.is_a?(TrueClass) ? "bigint" : exclude_output_inserted
                     <<~SQL.squish
-                      DECLARE @ssaIdInsertTable table (#{quoted_pk} #{id_sql_type});
-                      #{sql.dup.insert sql.index(/ (DEFAULT )?VALUES/i), " OUTPUT INSERTED.#{quoted_pk} INTO @ssaIdInsertTable"}
-                      SELECT CAST(#{quoted_pk} AS #{id_sql_type}) FROM @ssaIdInsertTable
+                      DECLARE @ssaIdInsertTable table (#{quoted_pk.map { |subkey| "#{subkey} #{id_sql_type}"}.join(", ") });
+                      #{sql.dup.insert sql.index(/ (DEFAULT )?VALUES/), " OUTPUT INSERTED.#{quoted_pk.join(', INSERTED.')} INTO @ssaIdInsertTable"}
```
aidanharan, line 286: Would you mind changing the regex to be case insensitive again? Also, refactored the code so that it uses a map and join like the lines above and below. 

```suggestion
                      #{sql.dup.insert sql.index(/ (DEFAULT )?VALUES/i), " OUTPUT #{ quoted_pk.map { |subkey| "INSERTED.#{subkey}" }.join(", ") } INTO @ssaIdInsertTable"}
```

jonhinson: Nice catch, done!

aidanharan:
@jonhinson Thanks for the fix!

jonhinson:
> @jonhinson Thanks for the fix!

You're welcome! Thanks for such a quick review!

aidanharan:
Fix in release [v7.1.3](https://rubygems.org/gems/activerecord-sqlserver-adapter/versions/7.1.3).

Issue #1163, opened 2024-04-23
exclude_output_inserted_table_names does not support composite primary keys with different data types


vividmuimui:
## Issue
The exclude_output_inserted_table_names feature does not handle composite primary keys with defferent data types properly.

## Expected behavior
The exclude_output_inserted_table_names should be able to handle composite primary keys of different data types without errors.

## Actual behavior
Currently, if you try to use exclude_output_inserted_table_names for a table with a composite primary key, this setting will not handle the scenario correctly. This limitation makes it particularly difficult to manage tables with composite keys containing different data types.

## How to reproduce

Create a table with a composite primary key where the keys have different data types and attempt to insert data using the ActiveRecord-SQLServer adapter with exclude_output_inserted_table_names configured. Here is an example setup that triggers the error:
```rb
create_table "SampleTable", primary_key: ["KeyPart1", "KeyPart2"], force: :cascade do |t|
  t.uuid "KeyPart1", null: false
  t.integer "KeyPart2", null: false # different data type
end
```

```rb
adapter = ActiveRecord::ConnectionAdapters::SQLServerAdapter
adapter.exclude_output_inserted_table_names[SampleTable] = 'uniqueidentifier'
```

Attempting to insert data into this table results in the following error:
```
  TRANSACTION (0.7ms)  BEGIN TRANSACTION
[...]

PR #1164, opened 2024-04-23
Fix composite primary key with different data type with triggers

vividmuimui:
This PR allows `adapter.exclude_output_inserted_table_names` to specify the type of each primary key when trigger is set and there are tables with composite primary keys.

```ruby
adapter.exclude_output_inserted_table_names['my_composite_pk_table_name'] = { pk_col_one: "uniqueidentifier", pk_col_two: "int" }
```

Fix https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1163


aidanharan:
Thanks for the PR!!!

PR #1166, opened 2024-05-02
Allow basic attribute types to be query parameters

aidanharan:
Previously when using place-holder conditions (`where("written_on < ?", time)`), the value was interpolated into the query (`WHERE written_on < "2024-01-01 12:00:00"`). Now the place-holder condition is a query parameter (`WHERE written_on < @1`).

Fixes the "undefined method 'value_for_database'" errors seen in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/actions/runs/8910633467/job/24470268020

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -331,11 +331,7 @@ module ActiveRecord
         def sp_executesql_sql_type(attr)
           return attr.type.sqlserver_type if attr.respond_to?(:type) && attr.type.respond_to?(:sqlserver_type)
 
-          value = if attr.is_a?(Symbol) || attr.is_a?(String) || attr.is_a?(Numeric)
-                    attr
-                  else
-                    attr.value_for_database
-                  end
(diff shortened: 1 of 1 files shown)


aaronsajo:
Hey @aidanharan ,
I am not sure if this is related, but we are still facing the same issue in Rails-7.2.1 and activerecord-sqlserver-adapter 7.2.1.
`.where('date_column < ?', date)` is throwing   "undefined method `value_for_database'"

andsip:
I also came across this today.

- `where("written_on < ?", Time.now)` ✅ 
- `where("written_on < ?", Date.today)` ❌ 
- `where("written_on < ?", DateTime.current)` ❌ 

So, it seems to work only for `Time` objects, but not `Date` or `DateTime` objects.

Adding checks for `type.is_a?(Date) || type.is_a?(DateTime)` in [`#basic_attribute_type?`](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/68d1c0f4c822323f032cdba0f2ec7844259238c3/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L361) would fix these specific cases, but I don't know the codebase well enough to say, if this would break something else.

PR #1198, opened 2024-07-02
Remove ActiveRecord::Relation#calculate patch

jonahgeorge:
Hey all, 

I've been working on adding SQL Server support to [ankane/groupdate](https://github.com/ankane/groupdate) and discovered that this patch of `ActiveRecord::Relation#calculate` is interfering with the [library's own patching](https://github.com/ankane/groupdate/blob/master/lib/groupdate/active_record.rb#L6) of [the same method](https://github.com/ankane/groupdate/blob/master/lib/groupdate/relation.rb#L11-L17).

It appears that all tests pass without this patch so posting this PR to potential remove it as I don't fully follow why it was introduced to start.


aidanharan:
The `ActiveRecord::Relation#calculate` patch was introduced by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1090 to fix an issue with failing `activerecord/test/cases/null_relation_test.rb` tests. Not sure exactly what tests now and since the tests pass now without the patch then all good 👍

The `main` branch is used for Rails 7.2 development. Do you need the patch removed from the 7.1 adapter?

jonahgeorge:
> The `ActiveRecord::Relation#calculate` patch was introduced by #1090 to fix an issue with failing `activerecord/test/cases/null_relation_test.rb` tests. Not sure exactly what tests now and since the tests pass now without the patch then all good 👍
> 
> @jonahgeorge The `main` branch is used for Rails 7.2 development. Do you need the patch removed from the 7.1 adapter?

Would you accept backports for both 7.1 and 7.0? Unfortunately the app that I really need this in is still on 7.0-- No worries if not

aidanharan:
Backports welcome 👍

aidanharan:
@jonahgeorge Released [v7.0.7](https://rubygems.org/gems/activerecord-sqlserver-adapter/versions/7.0.7) and [v7.1.4](https://rubygems.org/gems/activerecord-sqlserver-adapter/versions/7.1.4) with the back ports.

Issue #1207, opened 2024-07-26
TinyTds::Error: There is already an object named 'guest' in the database.


jvon1904:
## Issue
After release 7.1.5 which included #1201, you will receive an error when loading the schema because of the `create_schema` directive automatically added to `db/schema.rb`: 

```ruby
create_schema 'guest'
```

The "guest" schema is automatically created by MSSQL.  It's one of the  [four schemas that can't be dropped](https://learn.microsoft.com/en-us/sql/relational-databases/security/authentication-access/ownership-and-user-schema-separation?view=sql-server-ver16#built-in-schemas-for-backward-compatibility).

As such, it should be excluded along with `db_%`, `sys` and `INFORMATION_SCHEMA` from the SQL statement under `#schema_names` in `lib/active_record/connection_adapters/sqlserver/schema_statements.rb:403`.

```diff
# Returns an array of schema names.
        def schema_names
          sql = <<~SQL.squish
           SELECT name
            FROM sys.schemas
            WHERE
            name NOT LIKE 'db_%' AND
-          name NOT IN ('INFORMATION_SCHEMA', 'sys')
+          name NOT IN ('INFORMATION_SCHEMA', 'sys', 'guest')
          SQL

          query_values(sql, "SCHEMA")
        end
```

## How to reproduce
First migrate the database to update the `schema.rb` file.
```bash
$ bin/rails db:migrate
```

Next, load the schema in the Test environment, or run tests if `ActiveRecord::Migration.maintain_test_schema!` is set.

```bash
$ RAILS_ENV=test bin/rails db:test:load_schema --trace                                                                          19ms
** Invoke db:test:load_schema (first_time)
** Invoke db:test:purge (first_time)
** Invoke db:load_config (first_time)
** Invoke environment (first_time)
** Execute environment
** Execute db:load_config
** Invoke db:check_protected_environments (first_time)
** Invoke db:load_config
** Execute db:check_protected_environments
** Execute db:test:purge
** Execute db:test:load_schema
bin/rails aborted!
ActiveRecord::StatementInvalid: TinyTds::Error: There is already an object named 'guest' in the database. (ActiveRecord::StatementInvalid)
/Users/jvonbergen/.rvm/gems/ruby-3.3.1/gems/activerecord-sqlserver-adapter-7.1.5/lib/active_record/connection_adapters/sqlserver/database_statements.rb:445:in `do'
[...]

PR #1208, opened 2024-07-26
Exclude "guest" schmea in schema dumper

jvon1904:
This PR resolves the Issue #1207.

I added "guest" to the SQL query, and added a few extra lines of test code to check that the schema dumper does not add lines to create any of the MSSQL default schemas.


jvon1904:
FYI @aidanharan!  Thanks!

aidanharan:
@jvon1904 Thanks for the PR!

PR #1238, opened 2024-10-07
Allow INSERT statements with SELECT notation

### LINKED ISSUE #1237: Bug introduced by PR #1206 (allow spaces in table names)
## Issue
PR #1206 introduced support for spaces in table names.

## Expected behavior
Table names with spaces should be correctly escaped and interpreted.

For example:

- `INSERT INTO table name with spaces ...` should assume the table name is table and stop parsing after the first space.
- `INSERT INTO [table name with spaces] ...` should be the only syntax to treat spaces as part of the table name.

## Actual behavior
The method `ActiveRecord::ConnectionAdapters::SQLServer::SchemaStatements#get_raw_table_name` does not correctly identify table names in INSERT statements of the format `INSERT INTO tablename SELECT * FROM other_table`.

Currently, it returns the entire string `tablename SELECT * FROM other_table` as the table name, rather than stopping at tablename.

## How to reproduce
```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "tiny_tds"
  gem "activerecord", "7.2.1"
  gem "activerecord-sqlserver-adapter", "7.2.0"
end

require "active_record"
require "minitest/autorun"
require "logger"

  adapter:  "sqlserver",
  timeout:  5000,
  pool:     100,
  encoding: "utf8",
  database: "master",
  username: "SA",
  password: "StrongPassword!",
  host:     "127.0.0.1",
  port:     1433,
  )
ActiveRecord::Base.logger = Logger.new(STDOUT)

  drop_table :bug_tests rescue nil

  create_table :bug_tests, force: true do |t|
    t.bigint :external_id
  end
end

class BugTest < ActiveRecord::Base
end

class TestShouldBeWorking < Minitest::Test
  def setup
    BugTest.connection.execute("INSERT INTO bug_tests select('1')")
  end

  def test_should_be_working
    assert_equal 1, BugTest.count
  end
end

class TestWorking < Minitest::Test
  def setup
    BugTest.connection.execute("INSERT INTO bug_tests (external_id) VALUES ('1')")
  end

  def test_working
    assert_equal 1, BugTest.count
  end
end
```
## Details

- **Rails version**: `7.2.1`
- **SQL Server adapter version**: `7.2.0`
- **TinyTDS version**: `2.1.7`
- **FreeTDS details**:

  ```
                            Version: freetds v1.4.23
             freetds.conf directory: /etc/freetds
     MS db-lib source compatibility: yes
        Sybase binary compatibility: no
                      Thread safety: yes
                      iconv library: yes
                        TDS version: auto
                              iODBC: no
                           unixodbc: yes
              SSPI "trusted" logins: no
                           Kerberos: yes
                            OpenSSL: yes
                             GnuTLS: no
                               MARS: yes
  ```

zurchpet:
This Pull request fixes #1237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,5 +15,6 @@
 
 - [#1215](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1215) Fix mismatched foreign key errors
 - [#1228](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1228) Enable identity insert on view's base table
+- [#1238](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1238) Allow INSERT statements with SELECT notation
 
 Please check [7-1-stable](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/7-1-stable/CHANGELOG.md) for previous changes.
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -702,6 +702,7 @@ module ActiveRecord
             s.split(/INSERT INTO/i)[1]
               .split(/OUTPUT INSERTED/i)[0]
               .split(/(DEFAULT)?\s+VALUES/i)[0]
+              .split(/\bSELECT\b(?![^\[]*\])/i)[0]
               .match(/\s*([^(]*)/i)[0]
           elsif s.match?(/^\s*UPDATE\s+.*/i)
             s.match(/UPDATE\s+([^\(\s]+)\s*/i)[1]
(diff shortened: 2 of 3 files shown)


aidanharan:
@zurchpet The CI tests are failing. Tests will need to pass before I can merge the PR. 

The failing `ActiveRecord::PredicateBuilderTest` tests can be ignored as they have already been fixed in `main` by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1239

aidanharan:
@zurchpet Thanks for the PR. Do you need the change back ported to version 7.2?

zurchpet:
@aidanharan Thanks a lot!
Yes. Also in 7.1 actually.
We pin to version 7.1.4 for now.
Rails 7.2 or Rails 8.0 Upgrade is still in planing.

Issue #1241, opened 2024-10-10
Update example in section "Configure Connection" in README.md


andsip:
The following example code [is proposed at the moment](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/68d1c0f4c822323f032cdba0f2ec7844259238c3/README.md?plain=1#L101-L118):
```ruby
module ActiveRecord
  module ConnectionAdapters
    class SQLServerAdapter < AbstractAdapter
      def configure_connection
        super
        @raw_connection.execute("SET TEXTSIZE #{64.megabytes}").do
      end
    end
  end
end
```

However, as of commit 4842d2e, this implementation no longer works, as far as I see it.

This is because method `#configure_connection` was defined on `SQLServerAdapter` in that commit, which means that the example in the Readme will override this method, causing a call to `super` to invoke `AbstractAdapter#configure_connection` and not `SQLServerAdapter#configure_connection` (as intended).

The following adaptation of the example code works for me:

```ruby
ActiveRecord::ConnectionAdapters::SQLServerAdapter.prepend(
  Module.new do
    def configure_connection
      super
      @raw_connection.execute("SET TEXTSIZE #{64.megabytes}").do
    end
  end
)

```

aidanharan:
@andsip Good point! Could you review https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1242 and see what you think?

PR #1242, opened 2024-10-10
Updated sample code for connection configuration

aidanharan:
Custom connection configuration can now be performed by overriding the `additional_connection_configuration` method. Previous method as documented did not work. 

Addresses issue raised by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [#1196](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1196) Use default inspect for database adapter
 - [#1216](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1216) Refactor adapter interface to match abstract adapter
 - [#1225](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1225) Drop support to Ruby 3.1
+- [#1242](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1242) Changed how additional connection configuration is performed
 
 #### Fixed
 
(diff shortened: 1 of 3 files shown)


andsip:
Thanks for the quick response and PR @aidanharan.

I can confirm that your proposed change restores the behavior that existed prior to 4842d2e884346.

There is a use case though which I have, that cannot be solved by this.
That is, I need to adjust the `DATEFORMAT` setting like so:
```ruby
@raw_connection.execute("SET DATEFORMAT 'dmy'").do
```

Since this setting potentially influences [the setup in `initialize_dateformatter`](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1242/files#diff-3562da8bff63a7b06dce2a8b9e1163ab5598cca3bcaaf1b971a1c3e40942b053R522), I would need this setting to be run before `initialize_dateformatter`.
But with [`additional_connection_configuration` at the end](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1242/files#diff-3562da8bff63a7b06dce2a8b9e1163ab5598cca3bcaaf1b971a1c3e40942b053R524), it will run after.
But we also cannot put this method before in general, as e.g. for the `TEXTSIZE` setting, it needs to be run after.

That's why I proposed the approach using `#prepend` in #1241, as with this, the user has more flexibility by calling `super` before or after the configuration settings.
But maybe there are concerns about using this approach, which I am not aware of.

aidanharan:
Yeah, your approach is better as it gives the user more flexibility as you say. I've updated the README with your code.

PR #1246, opened 2024-10-15
Fix queries with date and date-time placeholder conditions

### LINKED ISSUE #1243: Issue with date values in queries
> I am not sure if this is related, but we are still facing the same issue in Rails-7.2.1 and activerecord-sqlserver-adapter 7.2.1.
> `.where('date_column < ?', date)` is throwing   "undefined method `value_for_database'"

_Originally posted by @aaronsajo in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1166#issuecomment-2404977768_
            

> I also came across this today.
> 
> where("written_on < ?", Time.now) ✅
> where("written_on < ?", Date.today) ❌
> where("written_on < ?", DateTime.current) ❌
> So, it seems to work only for Time objects, but not Date or DateTime objects.
> 
> Adding checks for type.is_a?(Date) || type.is_a?(DateTime) in [#basic_attribute_type?](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/68d1c0f4c822323f032cdba0f2ec7844259238c3/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L361) would fix these specific cases, but I don't know the codebase well enough to say, if this would break something else.

andsip:
Fix queries with date and date-time placeholder conditions such as:

```ruby
where("written_on < ?", Date.today)
where("written_on < ?", DateTime.current) 
```

This should fix #1243.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,5 +16,6 @@
 - [#1215](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1215) Fix mismatched foreign key errors
 - [#1228](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1228) Enable identity insert on view's base table
 - [#1238](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1238) Allow INSERT statements with SELECT notation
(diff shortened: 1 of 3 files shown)


aidanharan:
Could you add a few tests and update the `CHANGELOG`?

andsip:
I don't see any tests that have been added in the last several changes in this code part.. Could you point me to the test file, where similar functionality is being tested already?

Issue #473, opened 2016-05-16
Adding nonclustered Index with INCLUDE.


btagliani:
Hi everyone! I was wondering if there's any way to create an index like this: 
`CREATE NONCLUSTERED INDEX IndexName ON [dbo].[table] ([id],[type],[for_type],[attrib]) INCLUDE ([for_id],[value])` 
with ActiveRecord?

metaskills:
Just use the `execute` method in migrations when you need to do anything off the beaten path.

erlingur:
@metaskills This is now supported from Rails but when running this the index does not get correctly created with the INCLUDE. Which means it will not be supported in the schema.rb file which means you have to find a way to somehow run a separate script to set these covering indexes up after a new installation.

https://github.com/rails/rails/pull/44803

It would be great if this gem could support this. I have a performance issue in an app that I need to solve with covering indexes and having this built into the schema.rb file would be great.

erlingur:
@aidanharan  Thank you for reopening this! :) Hopefully this can get integrated.

aidanharan:
`INDEX INCLUDE` has been added to Rails 8.1 in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1301

erlingur:
Awesome, thank you so much @aidanharan !!

PR #1301, opened 2025-02-14
Add support for `INDEX INCLUDE`

aidanharan:
Add support for index include as requested in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/473

Running the following:

```ruby
connection.add_index("testings", "last_name", include: :foo)
```

Before this PR generates:

```SQL
CREATE INDEX [index_testings_on_last_name] ON [testings] ([last_name])
```

After the PR it will generate:

```SQL
CREATE INDEX [index_testings_on_last_name] ON [testings] ([last_name]) INCLUDE ([foo])
```

Multiple include columns are supported using:

```ruby
connection.add_index("testings", "last_name", include: [:foo, :bar])
```

Ref: https://github.com/rails/rails/pull/44803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+#### Added
+
+- [#1301](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1301) Add support for INDEX INCLUDE.
+
 #### Changed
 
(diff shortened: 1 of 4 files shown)


aidanharan:
Tests will be included as part of the `ActiveRecord` tests after https://github.com/rails/rails/pull/54532 is merged.

PR #1312, opened 2025-03-16
Support `insert_all` and `upsert_all` using `MERGE`

andyundso:
This PR adds support for `insert_all` and `upsert_all` in the `activerecord-sqlserver-adapter`. As initially propsed on #859 / #869, it uses `MERGE` to achieve so.

The benefit of `ON DUPLICATE` clauses in other database systems is that you can pass it every data that you have in your Rails app without the need for de-duplication. With `MERGE`, it technically executes a join between the data we want to insert and the data in the target table. therefore, the values in the columns of our source data, which will be joined, need to be unique. Unique means unique across all primary keys and indexes with uniqueness. The following Rails test show-cases this well:

```ruby
 def test_insert_all_with_skip_duplicates_and_autonumber_id_not_given
    skip unless supports_insert_on_duplicate_skip?

    assert_difference "Book.count", 1 do
      # These two books are duplicates according to an index on %i[author_id name]
      # but their IDs are not specified so they will be assigned different IDs
      # by autonumber. We will get an exception from MySQL if we attempt to skip
      # one of these records by assigning its ID.
      Book.insert_all [
        { author_id: 8, name: "Refactoring" },
        { author_id: 8, name: "Refactoring" }
      ]
    end
  end
```

From the adapter, we can ask Rails about indexes with uniqueness and all the primary keys for a given table. With that information, I run a `PARTITION BY` for each affected column (in the example above, one is the combination of `author_id` and `name`, the second is just the `id` column as primary key) over the source data, retrieve a `ROW NUMBER` and only keep the records where their `ROW NUMBER` is 1.

The rest of the code is then mostly as in other database adapters. More details are also provided in the code comments.

**Performance**

The big advantage of `insert_all` and `upsert_all` is performance. I did not run extensive benchmarks for now, but just tried a basic case, which is to insert 50'000 simple post records.

If I insert them all one by one ([gist](https://gist.github.com/andyundso/f63d3c651ec4d193bb82a4b37229bf14)), this takes about 144 seconds on my machine. If use `insert_all`, this takes 0.48 seconds. ([gist](https://gist.github.com/andyundso/665907d0257ded3f8fc2596672b2cf34)). If `insert_all` has a lot of collisions, like in [this gist](https://gist.github.com/andyundso/1fb249cd9bd20d12e70a568283f57f7f), the time goes up to 5 seconds. I also tested this collision script with MySQL, where the time to insert remains constant at 0.5 seconds.

I need to test some more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Added
 
 - [#1301](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1301) Add support for INDEX INCLUDE.
+- [#1312](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1312) Add support for `insert_all` and `upsert_all`
 
 #### Changed
 
--- a/README.md
+++ b/README.md
@@ -169,6 +169,22 @@ ActiveRecord::ConnectionAdapters::SQLServerAdapter.showplan_option = 'SHOWPLAN_X
 ```
 **NOTE:** The method we utilize to make SHOWPLANs work is very brittle to complex SQL. There is no getting around this as we have to deconstruct an already prepared statement for the sp_executesql method. If you find that explain breaks your app, simple disable it. Do not open a github issue unless you have a patch.  Please [consult the Rails guides](http://guides.rubyonrails.org/active_record_querying.html#running-explain) for more info.
 
+#### `insert_all` / `upsert_all` support
+
+`insert_all` and `upsert_all` on other database system like MySQL, SQlite or PostgreSQL use a clause with their `INSERT` statement to either skip duplicates (`ON DUPLICATE KEY IGNORE`) or to update the existing record (`ON DUPLICATE KEY UPDATE`). Microsoft SQL Server does not offer these clauses, so the support for these two options is implemented slightly different.
+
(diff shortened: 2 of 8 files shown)


Michoels:
Extremely exciting!
Kudos on this work, I can't wait to try it!

andyundso:
Performance for `upsert_all` looks good as well.

Upserting 50k records finishes in 2.07 seconds on my machine ([gist](https://gist.github.com/andyundso/06eef60468e6473e01a586f71b605cd1)). MySQL does does in 0.75 seconds ([gist](https://gist.github.com/andyundso/9769944be02ae9ca1335470ceb390cd2)). We likely loose some time because MSSQL can only process 1000 rows at the same time. this separation I do in Ruby, which likely is slow.

so I think this looks good overall and is ready for a review by @aidanharan.

aidanharan:
@andyundso Brilliant work! Will port the changes into the `8-0-stable` branch and get a v8.0 release out containing it. Will help with getting some early feedback.

andyundso:
@aidanharan I realised while driving home today that there is likely a bug with the order of our values when doing an `upsert`. I opened a PR against Rails to add tests for these scenarios, but it gives me conflicting results. Let's see if one of the maintainers have time to investigate it.

PR #1313, opened 2025-03-19
Correctly retrieve the SQL Server database version

aidanharan:
An exception was being silently thrown when getting the database version, and then defaulting to `2016`. 

In this PR:

- Fix the bug when retrieving the database version
- Stops the exception from being silently swallowed in future
- Outputs the database version after each test run to help debugging


bvogelzang:
This change looks like it would fix: https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1360. Specifically using `execute` instead of `_raw_select` to fetch the version information. Can we get this change ported back into `7-1`, `7-2` and `8-0`?

PR #1318, opened 2025-03-23
Reverse order of values when upserting

aidanharan:
Backport of https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+#### Fixed
+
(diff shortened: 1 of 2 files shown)


andyundso:
@aidanharan I am currently testing the `sqladapter` against the Solid Trifecta. So far, I have found at least one other bug with my implementation, but I am also not done yet with getting the test suites to pass. Maybe we could wait before issueing another release.

PR #1327, opened 2025-05-07
Fix multiple nil identity columns for merge insert

aidanharan:
Handle merge inserts that contain multiple `nil` identity columns. Eg:

```ruby
Book.upsert_all [
  { id: nil, name: "New edition 1" },
  { id: nil, name: "New edition 2" },
]
```

Previously the SQL generated would have been:

```sql
MERGE INTO [books] WITH (updlock, holdlock) AS target
using (SELECT *
       FROM   (SELECT [id],
                      [name],
                      [created_at],
                      [updated_at],
                      [updated_on],
                      Row_number()
                        OVER (
                          partition BY [id]
                          ORDER BY [id] DESC ) AS rn_0
               FROM   (VALUES 
                             (Ident_current(N'[books]') + Ident_incr(N'[books]'),
                      N'New edition 2',
                      CURRENT_TIMESTAMP,
                      CURRENT_TIMESTAMP,
                      CURRENT_TIMESTAMP),
                              (Ident_current(N'[books]') + Ident_incr(N'[books]'),
                      N'New edition 1',
                      CURRENT_TIMESTAMP,
                      CURRENT_TIMESTAMP,
                      CURRENT_TIMESTAMP)) AS t1 ([id], [name], [created_at],
                      [updated_at],
                      [updated_on])) AS ranked_source
       WHERE  rn_0 = 1) AS source
ON ( target.[id] = source.[id] )
WHEN matched THEN
  UPDATE SET target.[updated_at] = CASE
                                     WHEN (( source.[name] = target.[name]
                                              OR ( source.[name] IS NULL
                                                   AND target.[name] IS NULL ) )
                                          ) THEN
                                     target.[updated_at]
                                     ELSE CURRENT_TIMESTAMP
                                   END,
             target.[updated_on] = CASE
                                     WHEN (( source.[name] = target.[name]
                                              OR ( source.[name] IS NULL
                                                   AND target.[name] IS NULL ) )
                                          ) THEN
                                     target.[updated_on]
                                     ELSE CURRENT_TIMESTAMP
                                   END,
             target.[name] = source.[name]
WHEN NOT matched BY target THEN
  INSERT ([id],
          [name],
          [created_at],
          [updated_at],
          [updated_on])
  VALUES (source.[id],
          source.[name],
          source.[created_at],
          source.[updated_at],
          source.[updated_on])
output inserted.[id]; 
```

When this is run only the book with title "New edition 1" is created in the table. The reason is that "New edition 2" is first inserted but then "New edition 1" is updated over "New edition 2" since it matches `ON ( target.[id] = source.[id] )`. 

The fix changes the inserted ID columns to the following so that they don't match and both records get inserted:

- Ident_current(N'[books]') + (Ident_incr(N'[books]') * 1)
- Ident_current(N'[books]') + (Ident_incr(N'[books]') * 2)

This fixes the following tests which were added by https://github.com/rails/rails/pull/54962

- InsertAllTest#test_insert_all_implicitly_sets_primary_keys_when_nil
- InsertAllTest#test_upsert_all_implicitly_sets_primary_keys_when_nil

### DIFF
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -177,7 +177,7 @@ module ActiveRecord
               SELECT *
               FROM (
                 SELECT #{insert.send(:columns_list)}, #{partition_by_columns_with_uniqueness_constraints(columns_with_uniqueness_constraints:)}
-                FROM (#{insert.values_list})
+                FROM (#{merge_insert_values_list(insert:, insert_all:)})
                 AS t1 (#{insert.send(:columns_list)})
               ) AS ranked_source
               WHERE #{is_first_record_across_all_uniqueness_constraints(columns_with_uniqueness_constraints:)}
@@ -204,6 +204,35 @@ module ActiveRecord
           sql << ";"
 
           sql
+        end
+
+        # For `nil` identity columns we need to ensure that the values do not match so that they are all inserted.
+        # Method is a combination of `ActiveRecord::InsertAll#values_list` and `ActiveRecord::ConnectionAdapters::SQLServer::DatabaseStatements#default_insert_value`.
+        def merge_insert_values_list(insert:, insert_all:)
+          connection = insert.send(:connection)
+          identity_index = 0
+
+          types = insert.send(:extract_types_from_columns_on, insert.model.table_name, keys: insert.keys_including_timestamps)
+
+          values_list = insert_all.map_key_with_value do |key, value|
+            if Arel::Nodes::SqlLiteral === value
+              value
+            elsif insert.primary_keys.include?(key) && value.nil?
+              column = insert.send(:column_from_key, key)
+
+              if column.is_identity?
+                identity_index += 1
+                table_name = quote(quote_table_name(column.table_name))
+                Arel.sql("IDENT_CURRENT(#{table_name}) + IDENT_INCR(#{table_name}) * #{identity_index}")
+              else
+                connection.default_insert_value(column)
+              end
+            else
+              ActiveModel::Type::SerializeCastValue.serialize(type = types[key], type.cast(value))
+            end
+          end
+
+          connection.visitor.compile(Arel::Nodes::ValuesList.new(values_list))
         end
 
         # === SQLServer Specific ======================================== #
--- a/lib/active_record/connection_adapters/sqlserver/database_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/database_statements.rb
@@ -223,7 +223,7 @@ module ActiveRecord
               if column.is_identity?
                 identity_index += 1
                 table_name = quote(quote_table_name(column.table_name))
-                Arel.sql("IDENT_CURRENT(#{table_name}) + IDENT_INCR(#{table_name}) * #{identity_index}")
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -206,6 +206,35 @@ def build_sql_for_merge_insert(insert:, insert_all:, columns_with_uniqueness_con
...
+
+          types = insert.send(:extract_types_from_columns_on, insert.model.table_name, keys: insert.keys_including_timestamps)
+
+          values_list = insert_all.map_key_with_value do |key, value|
+            if Arel::Nodes::SqlLiteral === value
+              value
+            elsif insert.primary_keys.include?(key) && value.nil?
+              column = insert.send(:column_from_key, key)
+
+              if column.is_identity?
```
andyundso, line 223: just curious: is it possible under MSSQL that you have an identity column that is not a primary key? or is identity column just another word for "column with autogenerated identifier"?

lib/active_record/connection_adapters/sqlserver/database_statements.rb
```diff
@@ -206,6 +206,35 @@ def build_sql_for_merge_insert(insert:, insert_all:, columns_with_uniqueness_con
...
+                identity_index += 1
+                table_name = quote(quote_table_name(column.table_name))
+                Arel.sql("IDENT_CURRENT(#{table_name}) + (IDENT_INCR(#{table_name}) * #{identity_index})")
```
andyundso, line 226: shouldn't this be enough?

```suggestion
                Arel.sql("IDENT_CURRENT(#{table_name}) + (IDENT_INCR(#{table_name}) + #{identity_index})")
```

but I am not sure. I assume MSSQL uses the identifier generated here and does not overwrite them, correct? if it does overwrite it, then ignore this comment. if it does not overwrite it, then I would prefer the `+`, as it would generate much smaller values.

andyundso (approved):

lib/active_record/connection_adapters/sqlserver/database_statements.rb
aidanharan: The values given by `"IDENT_CURRENT(#{table_name}) + (IDENT_INCR(#{table_name}) + #{identity_index})"` are the IDs used when the record is created. I used multiplication as `IDENT_INCR` is the step value (https://learn.microsoft.com/en-us/sql/t-sql/functions/ident-incr-transact-sql?view=sql-server-ver16) that should be used for generating the ID. 

So if `IDENT_CURRENT('books)` is 100 and `IDENT_INCR('books')` is 2 then the next IDs generated should be 102, 104, 106, etc. To generate this sequence we need to use multiplication. If we used addition then the IDs generated would be 102, 103, 104, etc.

lib/active_record/connection_adapters/sqlserver/database_statements.rb
aidanharan: Had to look that up. Looks like you can have a primary key that is not an identity column and an identity column that is not a primary key.

```sql
CREATE TABLE ExampleTable (
    PrimaryKeyCol INT NOT NULL PRIMARY KEY,
    IdentityColumn INT IDENTITY(1,1) NOT NULL,
    SomeOtherCol NVARCHAR(100) NULL
);
```

Issue #1355, opened 2025-08-05
Error when performing SQL insert in cross database scenario.


chrisortman:
I believe this is the culprit. 

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/0adea5b20516661b510a7e571f3170152484fc16/lib/active_record/connection_adapters/sqlserver/database_statements.rb#L482

When you leverage the capability of SQL Server to operate across databases the lookup for identity columns is broken because schema_cache is bound to a single connection pool.

Just to illustrate with an example the following can be added to line 84 of `adapter_test_sqlserver.rb`

```
    sql = <<~SQL
    INSERT INTO #{arunit2_database}.dbo.colleges SELECT * FROM #{arunit_database}.dbo.topics
    SQL

    assert arunit_connection.send(:query_requires_identity_insert?, sql)
```

aidanharan:
Could you create a script to reproduce the issue? See https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/wiki/How-to-report-a-bug

chrisortman:
Yup, was working in it ....running into weird platform issues with sqlite3 gem. My RUBY_PLATFORM is aarch64-linux but the gem is locked to aarch64-linux-{gnu,musl} 

Going to try adding aarch64-linux to the lock file and see if that works. 

Problem is strange because _sometimes_ bundle install works and I'd expect it to always fail. But I had run the test suite before I submitted the bug report.

aidanharan:
Fixed

PR #1357, opened 2025-08-08
Support cross database inserts

aidanharan:
Related to https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1355

When finding the column information of a table, include the database when calculating the table's `object_id`. Otherwise the database is assumed to be the connection's default database.

Ref: https://learn.microsoft.com/en-us/sql/t-sql/functions/object-id-transact-sql?view=sql-server-ver17

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ jobs:
   test:
     name: Run test suite
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     env:
       COMPOSE_FILE: docker-compose.ci.yml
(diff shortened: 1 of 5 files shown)


chrisortman:
@aidanharan I think to make an eventually passing test the SQL needs to change to this https://github.com/chrisortman/activerecord-sqlserver-adapter/commit/bfb3cb982887e575a2bf622b00881785c3d52052

aidanharan:
@chrisortman: Would you mind reviewing the following PR? I tried to add a test that performed a `INSERT...SELECT...` cross database but it was causing the CI to hang (https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1357/commits/ee9b04239c12ef09b3c3ee7154e0b52f3e2525ca#diff-013de79fea2606a856806b513ba2f7b1def314278de4194c246d667f4537bf26R237) and I was not able to discover the root cause. I added tests similar to the ones you have in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1356.

chrisortman (approved):
This looks great. Thank you for working on this.

PR #1403, opened 2026-09-02
Fix insert_all/upsert_all for single character, temporary and three part table names

### LINKED ISSUE #1402: upsert_all raises NoMethodError when the table name is a single character
Hi there.

We hit this while upgrading an app to Rails 8, and I wanted to report it since the failure
is not obvious from the error. I used AI assistance to narrow down the cause and to check the
patch for regressions, but everything below was verified by running the script and the
adapter's test suite against a local SQL Server.

## Issue

`upsert_all` (and `insert_all` when it takes the `MERGE` path) raises `NoMethodError` when the
target table name is a single character. The same failure happens for temporary tables,
non-ASCII table names, names containing `$`, and three part (cross database) names.

## Expected behavior

`upsert_all` inserts the rows.

## Actual behavior

```
ActiveRecord::StatementInvalid: NoMethodError: undefined method `[]' for nil
    .../sqlserver/schema_statements.rb:723:in `get_raw_table_name'
    ...
    activerecord-8.0.5.1/lib/active_record/connection_adapters/abstract/database_statements.rb:177:in `exec_insert_all'
    activerecord-8.0.5.1/lib/active_record/insert_all.rb:54:in `execute'
    activerecord-8.0.5.1/lib/active_record/relation.rb:921:in `upsert_all'
```

## How to reproduce

Script (also fails on `8-0-stable` and `8-1-stable`):

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "tiny_tds"
  gem "activerecord", "~> 8.0.0"
  gem "activerecord-sqlserver-adapter", "~> 8.0.0"
  gem "minitest"
end

require "active_record"
require "minitest/autorun"
require "logger"

  adapter: "sqlserver",
  timeout: 5000,
  pool: 5,
  encoding: "utf8",
  database: ENV.fetch("DB_DATABASE", "activerecord_unittest"),
  username: ENV.fetch("DB_USERNAME", "rails"),
  password: ENV.fetch("DB_PASSWORD", ""),
  host: ENV.fetch("DB_HOST", "localhost"),
  port: ENV.fetch("DB_PORT", "1433").to_i
)

ActiveRecord::Base.logger = Logger.new($stdout, level: :error)

  create_table :x, force: true do |t|
    t.string :name
  end

  create_table :products, force: true do |t|
    t.string :name
  end
end

class SingleCharacterName < ActiveRecord::Base
  self.table_name = "x"
end

class Product < ActiveRecord::Base
end

class BugTest < Minitest::Test
  # Passes.
  def test_upsert_all_with_a_normal_table_name
    Product.upsert_all([{id: 1, name: "a"}])

    assert_equal 1, Product.count
  end

  # Fails with: NoMethodError: undefined method `[]' for nil
  def test_upsert_all_with_a_single_character_table_name
    SingleCharacterName.upsert_all([{id: 1, name: "a"}])

    assert_equal 1, SingleCharacterName.count
  end
end
```

Output:

```
2 runs, 1 assertions, 0 failures, 1 errors, 0 skips
```

The only difference between the two tests is the length of the table name.

## Cause

The `MERGE` branch of `get_raw_table_name` has two identifier slots, and both are `+`, so they
each need to match at least one character:

https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/cce2e38a025132c67b1711ebc59ce0352c100073/lib/active_record/connection_adapters/sqlserver/schema_statements.rb#L723

```ruby
s.match(/^\s*MERGE\s+INTO\s+(\[?[a-z0-9_ -]+\]?\.?\[?[a-z0-9_ -]+\]?)\s+(AS|WITH|USING)/i)[1]
```

`[products]` matches because the engine can split it across the two slots (`produc` + `ts`).
`[x]` cannot be split, so the match fails and `nil[1]` raises. The same applies to any name whose
characters are not in `[a-z0-9_ -]`, because the slot cannot consume them either.

Since the character class also allows a space and the second slot is reachable without a
preceding dot, an aliased target is swallowed into the captured name as well:

```ruby
# MERGE INTO [products] AS target USING ...
get_raw_table_name(sql)               # => "[products] AS target"
get_table_name(sql)                   # => "products] AS targe"
query_requires_identity_insert?(sql)  # => raises "Table '[products] AS target' doesn't exist"
```

This is not reachable from `build_sql_for_merge_insert` (it always emits
`WITH (UPDLOCK, HOLDLOCK) AS target`, and `WITH` terminates the match), but it is reachable from
hand written `MERGE` passed to `execute`.

## Affected cases

Checked against the released 8.0.11 gem:

| Table name | Current | With the fix below |
|---|---|---|
| `[x]` (single character) | `NoMethodError` | ok |
| `[#tmp]` (temporary table) | `NoMethodError` | ok |
| `[商品]` (non-ASCII) | `NoMethodError` | ok |
| `[my$tab]` | `NoMethodError` | ok |
| `[otherdb].[dbo].[dogs]` (three part) | `NoMethodError` | ok |
| `MERGE INTO [products] AS target USING ...` (hand written) | wrong name (see above) | ok |
| `MERGE INTO dbo .products ...` (space before the dot) | - | unchanged |

The aliased case does not raise inside `get_raw_table_name` itself; it returns a wrong name that
then fails downstream, as shown in the Cause section above.

`INSERT`, `UPDATE` and `FROM` already handle three part names; only the `MERGE` branch is limited
to two.

## Fix

I have a patch ready and can open a PR if you would like it:

```ruby
MERGE_TARGET_IDENTIFIER = /(?:\[[^\]]+\]|[a-z0-9_-]+)/i
MERGE_TARGET_TABLE_NAME = /\A\s*MERGE\s+INTO\s+(#{MERGE_TARGET_IDENTIFIER}(?:\s*\.#{MERGE_TARGET_IDENTIFIER}){0,2})\s+(?:AS|WITH|USING)/i
```

Spaces are allowed only inside brackets, the schema qualifier requires a dot, and up to three
parts are accepted (`database.schema.table`; a four part remote name is not a valid `MERGE`
target). Running the script above against this patch gives `2 runs, 2 assertions, 0 failures,
0 errors`, and `rake test ONLY_SQLSERVER=1` shows no change in failures or errors compared to
`main` (376 -> 384 runs, same 1 failure / 30 errors, which are pre-existing on `main`).

The `MERGE` line is byte identical on `main`, `8-0-stable` and `8-1-stable`, so the same patch
applies to all three.

## Details

- **Rails version**: `8.0.5.1`
- **SQL Server adapter version**: `8.0.11`
- **TinyTDS version**: `3.4.0`
- **FreeTDS details**:

  ```
  Version: freetds v1.4.26
  freetds.conf directory: /opt/homebrew/etc
  Thread safety: yes
  iconv library: yes
  TDS version: 7.3
  iODBC: no
  unixodbc: yes
  SSPI "trusted" logins: no
  Kerberos: yes
  ```

- **SQL Server**: `Microsoft SQL Server 2019 (RTM-CU32-GDR) 15.0.4455.2 (X64) on Linux (Ubuntu 20.04.6 LTS)`

Alt70155:
Fixes #1402.

I used AI assistance to narrow down the cause and to check the patch for
regressions, but everything here was verified by running the reproduction script
and the adapter's test suite against a local SQL Server 2019.

`upsert_all` (and `insert_all` when it takes the `MERGE` path) raises
`NoMethodError` when the `MERGE` table name pattern cannot match the target, and
returns a wrong name when the target carries an alias. See #1402 for the full
analysis of why the two identifier slots behave that way.

`INSERT`, `UPDATE` and `FROM` already handle three part names; only `MERGE` was
limited to two.

## Fix

Replace the pattern with an explicit identifier that matches either a bracket
quoted or a bare name, joined by up to two dots:

```ruby
MERGE_TARGET_IDENTIFIER = /(?:\[[^\]]+\]|[a-z0-9_-]+)/i # :nodoc:
MERGE_TARGET_TABLE_NAME = /\A\s*MERGE\s+INTO\s+(#{MERGE_TARGET_IDENTIFIER}(?:\s*\.#{MERGE_TARGET_IDENTIFIER}){0,2})\s+(?:AS|WITH|USING)/i # :nodoc:
```

Notes on the choices:

- A bracket quoted segment is `\[[^\]]+\]`, so spaces inside brackets are kept
  (`[WITH - SPACES]`) while a space is no longer allowed in a bare name. That is
  what stops an aliased target being captured as part of the table name.
- The dot separator is `\s*\.`, not `\.`, so `MERGE INTO dbo .products` keeps
  working. It parses today and I did not want to change that.
- `{0,2}` allows one, two and three part names, matching the other branches.

## Tests

- `test/cases/schema_test_sqlserver.rb` — 8 cases added to the existing
  `MERGE statements` block (5 -> 13): bracket quoted and bare names, a single
  character name, a name with spaces inside brackets, two and three part names,
  with and without a `WITH (UPDLOCK, HOLDLOCK)` hint, and the space before the
  dot case.
- `test/cases/adapter_test_sqlserver.rb` — one more assertion in the
  `query_requires_identity_insert?` test, using the generated `MERGE` with the
  table hint removed so the aliased target is exercised.

Without the change these fail with 3 failures and 2 errors; with it the file is
green.

```
schema_test_sqlserver.rb     32 runs, 42 assertions, 0 failures, 0 errors
adapter_test_sqlserver.rb    83 runs, 187 assertions, 0 failures, 0 errors, 1 skip
insert_all_test_sqlserver.rb  1 runs,  1 assertions, 0 failures, 0 errors
```

`bundle exec standardrb` is clean for the whole repository.

## Branches

The same pattern is present on `main`, `8-1-stable` and `8-0-stable`. This PR
targets `main`. Happy to open backport PRs for the stable branches if you would
like them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,5 +9,6 @@
 - [#1381](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1381) Fix `change_column` to preserve old column attributes.
 - [#1393](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1393) Treat TinyTDS `failed dbsqlsend() function` errors as `ConnectionNotEstablished` so they're retried instead of surfaced as `StatementInvalid`.
 - [#1397](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1397) Treat "DBPROCESS is dead or not enabled" as ConnectionNotEstablished.
+- [#1403](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1403) Fix `insert_all`/`upsert_all` for single character, temporary, non-ASCII and three part table names, and stop an aliased target being included in the `MERGE` table name.
 
 Please check [8-1-stable](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/blob/8-1-stable/CHANGELOG.md) for previous changes.
--- a/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
+++ b/lib/active_record/connection_adapters/sqlserver/schema_statements.rb
@@ -4,6 +4,9 @@ module ActiveRecord
   module ConnectionAdapters
     module SQLServer
       module SchemaStatements
+        MERGE_TARGET_IDENTIFIER = /(?:\[[^\]]+\]|[a-z0-9_-]+)/i # :nodoc:
+        MERGE_TARGET_TABLE_NAME = /\A\s*MERGE\s+INTO\s+(#{MERGE_TARGET_IDENTIFIER}(?:\s*\.#{MERGE_TARGET_IDENTIFIER}){0,2})\s+(?:AS|WITH|USING)/i # :nodoc:
+
         def create_table(table_name, **options)
           res = super
           clear_cache!
@@ -730,7 +733,7 @@ module ActiveRecord
           elsif s.match?(/^\s*UPDATE\s+.*/i)
             s.match(/UPDATE\s+([^(\s]+)\s*/i)[1]
           elsif s.match?(/^\s*MERGE INTO.*/i)
-            s.match(/^\s*MERGE\s+INTO\s+(\[?[a-z0-9_ -]+\]?\.?\[?[a-z0-9_ -]+\]?)\s+(AS|WITH|USING)/i)[1]
+            s.match(MERGE_TARGET_TABLE_NAME)[1]
           else
             s.match(/FROM[\s|(]+((\[[^(\]]+\])|[^(\s]+)\s*/i)[1]
           end.strip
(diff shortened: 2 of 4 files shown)


aidanharan:
@Alt70155 Could you create a PR against the `8-1-stable` branch instead of `main`? The tests aren't passing on `main` at the moment so cannot tell if your changes introduce any regressions. Thanks

Alt70155:
Thanks for taking a look. Opened #1405 against `8-1-stable`.

Issue #1396, opened 2026-08-18
translate_exception never maps "DBPROCESS is dead or not enabled" to a connection error


bhenn:
---

## Versions

- `activerecord-sqlserver-adapter` 8.0.5
- `tiny_tds` 2.1.7
- `rails` 8.0.2
- FreeTDS 1.4.26
- SQL Server (on-prem, TLS-encrypted TDS connection)

## Summary

When the TCP/TLS connection underlying a `TinyTds::Client` is killed while **idle** (no in-flight query), `tiny_tds`/FreeTDS has no way to know until the *next* query is attempted on it. That query fails with:

```
TinyTds::Error: DBPROCESS is dead or not enabled
```

`SQLServerAdapter#translate_exception` doesn't match this message against any of its patterns, so it falls through to `StatementInvalid` — a plain "your SQL was bad" error, not a connection error. This has two compounding effects, one already known, one that as far as we can tell is new:

### 1. (Previously reported, still unfixed) No reconnect happens
Because the exception isn't `ConnectionNotEstablished`, Rails has no signal that the connection itself is broken. The pool checks the same dead connection back in, and the next request that happens to check it out fails identically. This repeats until the process is restarted. This exact symptom has been reported and closed without a landed fix at least 9 times since 2013: #267, #402, #510, #525, #623, #668, #970 (and probably others we didn't find).

### 2. (New, as far as we can tell) Rails 8's own self-healing mechanism can't save this either
Rails 8's `AbstractAdapter#with_raw_connection` added exactly the kind of resiliency this needs — it marks a connection `@verified = false` after any non-retryable query error, and calls `verify!` (which calls `active?`) before the connection's next use, reconnecting if it's not active. On paper this should already fix the cascade, no adapter change needed.

In practice it doesn't help here, because:

```ruby
# tiny_tds ext/tiny_tds/client.c
static VALUE rb_client_dead_p(VALUE self) {
  ...
  return dbdead(cwrap->client) ? Qtrue : Qfalse;
}
```

[...]

PR #1412, opened 2026-09-16
Use TinyTDS Client#ping in active? when available

emailrhoads:
## Summary
- When the TinyTDS client responds to `#ping`, `SQLServerAdapter#active?` uses a FreeTDS-bounded liveness round-trip (`:ping_timeout`, default 2 seconds) instead of relying only on passive `dbdead` / `active?`.
- Set `:ping_timeout` to `0` to disable the checkout ping and keep the passive path.
- Falls back to today’s passive `@raw_connection.active?` on TinyTDS builds without `#ping`.

Depends on / follows [tiny_tds#609](https://github.com/rails-sqlserver/tiny_tds/pull/609), which adds `Client#ping(timeout:)` + `query_timeout=`. This is the adapter-side policy for the checkout-liveness discussion in [#1396](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1396): the driver owns the short-timeout probe; the adapter chooses to call it from `active?` (used by Rails 8 `verify!`).

This adapter used to do a live wellness check before [#1121](https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/pull/1121) (Oct 2023, “Removed legacy connection handling”), `active?` ran `raw_connection_do "SELECT 1"`. That PR switched to TinyTDS’s passive `active?` / FreeTDS `dbdead`, which only flips after a failed I/O and therefore cannot detect an idle TCP RST (the peer closes the socket while the client still thinks the session is fine).

This restores checkout liveness in the spirit of the pre-#1121 `SELECT 1` and of other adapters, without hanging on the connection’s timeout value, which may be much larger than is appropriate for this simple ping.


emailrhoads:
Tracking issue: https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/1413

